PR #20, opened 2015-08-24
remove enchanter addon from magi build

bbailey:
The enchanter is included by default so it's not necessary to add it to magi.


vadi2:
The version change shouldn't be in the PR, that's done at release only - could you fix that?

PR #26, opened 2015-08-24
Add installation instructions for OSX.

bunster:
Added the Mac instructions for installation and usage.


proram:
cloned svof out of github and was going through the rest of the setup instructions for mac, when tried to brew install lua5.1 got error that there was no available formula to lua5.1. 

Not familiar with homebrew - is there some other repo that needs to be setup?

Issue #3, opened 2015-08-22
Make Svof reloadable


vadi2:
Right now if you make a change in the core Lua files, you have to restart Mudlet in order to get the new changes in. It should be possible to keep the state and reload the system on the fly to make it easier to edit.

vadi2:
I'm adding 100 bounty credits to this.

QwindorRousseau:
@vadi2 does this still need to be open with the new system?

vadi2:
Nope - already did this myself

PR #27, opened 2015-08-24
Make Svof reloadable.

macjabeth:
Using the **vreload** alias, it should now be possible to keep the state
and reload the system on the fly to make it easier to edit.

### DIFF
--- a/svo (install the zip, not me).xml
+++ b/svo (install the zip, not me).xml
@@ -52773,6 +52773,13 @@ send(&quot;ir&quot;)</script>
                 <packageName></packageName>
                 <regex>^vinstall( fresh)?$</regex>
             </Alias>
+            <Alias isActive="yes" isFolder="no">
+                <name>Reload system</name>
+                <script>svo_reload()</script>
+                <command></command>
+                <packageName></packageName>
+                <regex>^vreload$</regex>
+            </Alias>
             <AliasGroup isActive="yes" isFolder="yes">
                 <name>Parry config</name>
                 <script></script>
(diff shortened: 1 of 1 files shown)


vadi2:
Where's the keeping the state bit?

bunster:
What do you mean by "keeping the state"? What state is being kept?

vadi2:
Well the description says "it should now be possible to keep the state", which state is it talking about?

bunster:
Keeping the current profile loaded / being able to reload the system.

vadi2:
The system state needs to be kept though, as not everything will initialise the same way out of the blue - it relies on the login messages and GMCP data. For example the character name will be unknown, and that will break some things.

bunster:
I see... Been rattling my brain on an answer to this. One idea could be keeping a list of which files are important to keep synchronised, so before you reload the system, you could save them into another table, and then merge them back in once the system is reloaded. Thoughts?

vadi2:
A general design like that would work, the trick is loading everything back right and initialising things that need initialising.

bunster:
Would something like this work?
```lua
function svo_reload()
  local attrs = {}
  local save_keys = {}
  -- Possibly, if we make a list of important keys:
  --[=[for i = 1, #save_keys do attrs[save_keys[i]] = package.loaded.svo[save_keys[i]] end --]=]
  for key, value in pairs(package.loaded.svo) do
    attrs[key] = value
  end
  package.loaded.svo = nil
  -- loader for packages
  local foldername
  for item in lfs.dir(getMudletHomeDir()) do
    foldername = item:match("(%w+) svo")
    if foldername and lfs.attributes(getMudletHomeDir().."/"..foldername.." svo/svo") then
      local oldpath = package.path
      local system_path = getMudletHomeDir().."/"..foldername.." svo/"
      package.path = system_path .. "?;" .. system_path .. "?.lua;" .. system_path .. "?.dll;" .. system_path .. "?.so;"

      local s,m = pcall(require, "svo")
      package.path = oldpath

      if not s then display(m) return end

      -- The issue would be if we changed/updated something in one of the already defined global tables.
      for key, value in pairs(attrs) do
        svo[key] = value
      end

      svo.echofn("Reloaded & ready to go, version %s (", tostring(svo.version))
      echoLink("homepage", '(openUrl or openURL)"http://svo.vadisystems.com"', "Svo homepage")
      decho(svo.getDefaultColor()..")\n")
      return
    end
  end
end
```

vadi2:
Test it out - reload the system, then run with it for a while - see if
anything breaks.

vadi2:
By the looks of the above you'll be overwriting all new functions with the
old ones, kind of negating reloading.

bunster:
Well sadly that is the only way of doing it, hence adding the save_keys table commented out. If we can make a list (I'm not sure which ones should be sync'd), then we can uncomment out that method instead.

vadi2:
Yep, it's not trivial, that's why it's not in yet as a feature.

keneanung:
Don't mind me, only marking connection wth #3

keneanung:
@bunster any updates on this?

vadi2:
Closing it in favour of https://github.com/svof/svof/pull/329, which makes Svof reloadable without losing the system state.

PR #33, opened 2015-08-24
Cadmus Wears Off

alex-d-schroder:
Domination Cadmus has a line when it wears off after 20 seconds. No trigger in the system for it other than a temp trigger, and that called a function not even implemented in the system. I added the trigger for Cadmus wearing off, and implemented the function. Based it off of similar functions for afflictions that wear off on their own.

### DIFF
--- a/raw-svo.valid.main.lua
+++ b/raw-svo.valid.main.lua
@@ -1528,6 +1528,11 @@ function valid.ablaze_woreoff()
   lifevision.add(actions.ablaze_gone.p)
 end
 
+function valid.cadmus_woreof()
+	checkaction(dict.cadmus.gone, true)
+	lifevision.add(actions.cadmus_gone.p)
+end
+
 function valid.wake_start()
   checkaction(dict.sleep.misc)
   if actions.sleep_misc then
--- a/raw-svo.valid.main.lua
+++ b/raw-svo.valid.main.lua
@@ -1528,11 +1528,6 @@ function valid.ablaze_woreoff()
   lifevision.add(actions.ablaze_gone.p)
 end
 
-function valid.cadmus_woreof()
-	checkaction(dict.cadmus.gone, true)
-	lifevision.add(actions.cadmus_gone.p)
-end
-
 function valid.wake_start()
   checkaction(dict.sleep.misc)
   if actions.sleep_misc then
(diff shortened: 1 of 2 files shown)


vadi2:
Have you tested this not working before?

The temp trigger only needs to exist for one line, and it does call a function that is implemented in the system:

```
for _, aff in ipairs({"hamstring", "galed", "voided", "inquisition", "burning", "icing", "phlogistication", "vitrification", "corrupted", "mucous", "rixil", "palpatar", "cadmus", "hecate", "ninkharsag", "swellskin", "pinshot", "dehydrated", "timeflux", "lullaby", "numbedleftarm", "numbedrightarm", "unconsciousness",
#if skills.metamorphosis then
"cantmorph",
#end
}) do
  valid[aff.."_woreoff"] = function()
    checkaction(dict[aff].waitingfor, true)
    if actions[aff.."_waitingfor"] then
      lifevision.add(actions[aff.."_waitingfor"].p)
    end
  end
end
```

alex-d-schroder:
I didn't see the implementation (I'm still learning my way around the system, mostly by doing these additions as best I can), but yeah, I can see that's not needed now.

However, I tested before, and again just now, and when the line showing the curse wearing off does not remove the affliction. Interestingly, the test revealed that Rixil wearing off doesn't remove its affliction either. So, that's going to have to be looked at, too.

vadi2:
I did vaff cadmus, then called svo.valid.cadmus_woreoff() and the affliction was removed:

```
3750h, 4730m, 21900e, 20400w exd-vaff cadmus
3750h, 4730m, 21900e, 20400w exd[cadmus]-
3750h, 4730m, 21900e, 20400w exd[cadmus]-vlua svo.valid.cadmus_woreoff()
(svof): Code ran OK, there were no results.
3750h, 4730m, 21900e, 20400w exd[cadmus]- (6.771s)
3750h, 4730m, 21900e, 20400w exd-
```

alex-d-schroder:
Yes, the function works, but it's never getting called. I had Seragorn afflict me with cadmus, no pausing of the system, and when cadmus wore off, the system still thought I had it until I diagnosed.

vadi2:
They did do some changes to it in the recent announce, so perhaps the way
it's delivered has changed. A log would help show.

PR #34, opened 2015-08-25
Changed two-handed focus trigger to use the precision line.

Seifur:
Allows for the proper adding of double stacks when hit with a precision focused attack.

### DIFF
--- a/svo (install the zip, not me).xml
+++ b/svo (install the zip, not me).xml
@@ -48965,7 +48965,7 @@ svo.valid.simplecrippledleftleg()</script>
                         <colorTriggerFgColor>#000000</colorTriggerFgColor>
                         <colorTriggerBgColor>#000000</colorTriggerBgColor>
                         <regexCodeList>
-                            <string>^(\w+) shifts into a more aggressive stance\.$</string>
+                            <string>^An expression of calculated focus crosses the face of (\w+), and s?he shifts h(?:er|is) feet slightly wider apart\.$</string>
                         </regexCodeList>
                         <regexCodePropertyList>
                             <integer>1</integer>


vadi2:
What about the aggressive stance, is that no longer a thing?

Seifur:
That's focus speed. It only applies one stack per hit.

vadi2:
Could you add a log of diagnose before that line and a diagnose after that line?

Seifur:
I'm just going to close this. Did my own testing this time and no change is necessary. My "helper" earlier told me he had used speed first and then precision. Turns out it was the reverse. I should have just diagnosed instead of taking his word for it.

```
(svo): System paused.
curing off
diagnose
diag
You are:
blind.
deaf.
(p) 10168h, 7566m, 46320e, 34560w xcdbk 57.2-(+19m, 0.3%)
You have recovered equilibrium. (0.975s)
(p) 10168h, 7566m, 46320e, 34560w excdbk 57.2-'ok
Seifur shifts into a more aggressive stance.
Seifur comes around with a terrible swing of a colossal, rune-etched warhammer toward your left leg.
Your leg almost gives out as the hammer crunches brutally into your ankle.
You gasp as your fine-tuned reflexes disappear into a haze of confusion.
(p) 9751h, 7547m, 46320e, 34559w excdbk 57.2-(-417h, 4.1%, -19m, 0.3%)
diagnose
diag
You are:
blind.
deaf.
suffering from 2 torn tendons.
afflicted with clumsiness.
bleeding for 9 health.
(p) 9751h, 7547m, 46320e, 34559w xcdbk 57.2-
```

vadi2:
Now you understand why I've always asked for logs :) thanks for taking the time to look at this

PR #35, opened 2015-08-25
Refiller toxicology addition

alex-d-schroder:
Added in the necessary stuff for using the refiller module to do
Toxicology fills. Added a variable so that transcendence of both skills
could be used without duplicate functions.

I haven't been able to test this myself, unfortunately. Needs someone
with remedies and toxicology (preferably both on the same character) to
give a go.

Other minor, non-toxicology changes made:

Removed a reference to the Concoctions skill (it was present during the
forestal skill transition, and never removed).

The caloric ingredient list was different than the others; I changed it
for more consistency; there should be no functional changes.

### DIFF
--- a/raw-svo.refiller.lua
+++ b/raw-svo.refiller.lua
@@ -15,6 +15,9 @@ rf_refilling = false
 -- format: "potion"
 -- this stores the current potion that we're doing of the whole order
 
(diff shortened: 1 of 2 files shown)


vadi2:
Thanks for the contribution!

PR #45, opened 2015-08-29
Add Travis config to generate a release and docs

### LINKED ISSUE #36: Setup documentation on Github Pages
http://doc.svo.vadisystems.com will be retired and the documentation needs to be hosted on Github instead.

keneanung:
The travis.yml will not work after commit into the svof repo as each encryption
key is bound to each repo. My preference would be to create a machine account
and record the credentials in a special helpfile in the svof clan with read
permissions for a special clan position which corresponds to the owner group in
the org. I'd be willing to set everything up with that bot.

This fixes #36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: c
+sudo: required
+install:
+- sudo apt-get update
+- sudo apt-get install lua5.1 luarocks p7zip-full python-sphinx
+- sudo luarocks install penlight
+script:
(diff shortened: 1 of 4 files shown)


vadi2:
Great work, thanks. At the moment there isn't anybody else in the svof org, as I'm giving it a bit of time until there are people available who know svof and its code well enough, so create the machine account and message me the details privately and I'll write it up.

PR #54, opened 2015-09-01
Generate release changelog from PRs and Issues

keneanung:
An example can be seen at https://github.com/keneanung/svof/releases/tag/14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: c
+language: ruby
 sudo: required
 install:
 - sudo apt-get update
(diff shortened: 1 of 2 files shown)


keneanung:
With the current setup, this is hard to test, but I am confident that this works. We could create a test-release after the merge and delete it afterwards.

keneanung:
This is now good to go.

If you add the "bug" or "enhancement" label to PRs, they get sorted into one of these categories instead of general "Merged PRs".

We also should think about adding a Contributing file (https://help.github.com/articles/setting-guidelines-for-repository-contributors/) to stress that the PR titles should be meaningful.

Issue #50, opened 2015-08-31
Refiller doesn't detect transcendent in Remedies


slester:
Despite having transcendent remedies, the refiller still tries to put 5 pours instead of 4 into each vial, which means I have to refill by hand (I get the "that would overflow your container" error).

vadi2:
@alex-d-schroder want to have a look?

alex-d-schroder:
I'll go through and figure out how it broke. It might not be until tomorrow (Sept/3/15) that I'll have the time to get a fix in.

PR #56, opened 2015-09-03
Refiller fix

alex-d-schroder:
Issue #50 Referred to the currenttrans variable incorrectly as svo.currenttrans when pouring into a vial, causing it to always resolve as false and pouring too much for transcendent brewers. Affected both standard and artie vials. Variables have been fixed, and it's filling vials correctly now.


vadi2:
That should still work the same as the refiller is loaded as part of the Svo module, but if that's the fix then so be it!

PR #70, opened 2015-09-15
Fix changelog generation by travis

keneanung:
Could you make a test release after the merge? I'll let you know when I checked the result and the release can be deleted again.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -8,6 +8,7 @@ fi
 
 lua precommit.lua "$TRAVIS_TAG"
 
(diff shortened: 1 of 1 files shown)


vadi2:
I've created the test release.

keneanung:
Hrm, looks like I didn't fix everything, I hoped the other was an error following, but it doesn't seem to be like that... :-1:

vadi2:
For some reason I can't delete the release either. The button doesn't show.

PR #71, opened 2015-09-15
More fixes to the travis changelog generation

keneanung:
Looks like I forgot some subshell and variable interpolation. I could have been retty tired when doing this...

Could you do another test release?


vadi2:
selection_020

Issue #40, opened 2015-08-25
Separate priority lists for affs/defs on serverside


vadi2:
Serverside actually has 2 queues - one for afflictions, and one for defences. This isn't immediately obvious in documentation (and puts into question of what will happen if an affliction and a defence share a balance), but Svof right now treats them as in the same queue. Svof's feature of syncing priorities to serverside needs to split up affs and defs into two separate lists before updating them on serverside.

QwindorRousseau:
As an addon to this is it already possible to set different prios per def profile so you could easily sway prios depending on who your fighting along with different health mana and moss percentages per profile without having to extensively code in auto prio swaps for your defence setup when you change to that defence setup

keneanung:
@vadi2 I'm not sure what to do with this? #87 split afflictions and defences into 2 tables and people seem to be happier now. Is this closed with the mentioned PR?

vadi2:
Okay, if it works, it works

PR #87, opened 2015-10-14
Count priorities for server side differently

keneanung:
Previously, gaps were closed when calculating priorities for server side. This
caused blindness and deafness defenses to be very high in the defence list even
though they appeared low in 'vp herbs'. By keeping those gaps, defences don't
get raised implicitely, but more defences pool at priority 25 as well.

### DIFF
--- a/raw-svo.controllers.lua
+++ b/raw-svo.controllers.lua
@@ -925,8 +925,8 @@ function printorder(balance, limited_around)
   local list = prio.getsortedlist(balance)
   local affs, defs = sk.splitdefs(balance, list)
   local raffs, rdefs = {}, {}
-  for i = 1, #affs do raffs[affs[i]] = i end
-  for i = 1, #defs do rdefs[defs[i]] = i end
+  for i = 1, table.maxn(affs) do if affs[i] then raffs[affs[i]] = i end end
+  for i = 1, table.maxn(defs) do if defs[i] then rdefs[defs[i]] = i end end
 
   if limited_around then
     echofn("(")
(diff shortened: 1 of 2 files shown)


vadi2:
The issue isn't the gaps - it's that afflictions and defences are actually on different priorities serverside, and Svof thinks they are on the same priority list. See https://github.com/svof/svof/issues/40. I don't think this fix is correct, fixing #40 would be more proper.

keneanung:
From my experience, both lists are practically merged on checking for
actions to do. That is defence prio 3 is the same as affliction prio 3. The
different lists only exist so defences and affluctions are not lumped
together.

I had Dochitha fight with those changes and she cured better this way.

Vadim Peretokin <notifications@github.com> schrieb am Mi., 14. Okt. 2015
23:34:

keneanung:
To add to this, I believe the issue stems from the fact that svof _believes_ that afflictions and defs are different queues and thus counts both from priority one. This leads to Blindness and deafbess being on prio 2 and 3 and thus before most herbal affliction cures.

vadimnehta:
Yep you're right, they need to be mixed in together.

keneanung:
And that is what this changes does: it makes szre the priorities stay the same, no matter whether it's a defence or affliction.

raw-svo.controllers.lua
```diff
@@ -925,8 +925,8 @@ function printorder(balance, limited_around)
   local list = prio.getsortedlist(balance)
   local affs, defs = sk.splitdefs(balance, list)
   local raffs, rdefs = {}, {}
-  for i = 1, #affs do raffs[affs[i]] = i end
-  for i = 1, #defs do rdefs[defs[i]] = i end
+  for i = 1, table.maxn(affs) do if affs[i] then raffs[affs[i]] = i end end
```
vadi2: table.maxn is depricated (http://www.lua.org/manual/5.2/manual.html).

keneanung:
It's using the pairs iterator now, which should be more efficient as well.

vadi2:
Not more efficient, but correct!

+1

PR #88, opened 2015-10-14
Different ndb fixes

### LINKED ISSUE #38: Update generate -o to also save ndb help file
The ndb-help.lua help file needs to be generated into the "own svo" folder as well as into the svo template folder, so developer version of the system can have a working "ndb" alias.

keneanung:
As the title says.... See commits for details. This also fixes #38


vadi2:
+1

PR #89, opened 2015-10-14
Make multiple enchantments in one command possible

keneanung:
You can now select a template item when enchanting multiple items with the same short description with the same enchantment. To enable this, a full inventory tracker is added to svo.

### DIFF
--- a/raw-svo.setup.lua
+++ b/raw-svo.setup.lua
@@ -188,11 +188,43 @@ end)
 signals.gmcproominfo        = luanotify.signal.new()
 signals.gmcpcharstatus      = luanotify.signal.new()
 signals.gmcpcharitemslist   = luanotify.signal.new()
+signals.gmcpcharitemslist:connect(function()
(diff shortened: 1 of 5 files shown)


raw-svo.setup.lua
```diff
@@ -188,11 +188,43 @@ end)
 signals.gmcproominfo        = luanotify.signal.new()
 signals.gmcpcharstatus      = luanotify.signal.new()
 signals.gmcpcharitemslist   = luanotify.signal.new()
+signals.gmcpcharitemslist:connect(function()
+  if not gmcp.Char.Items.List.location then echof("(GMCP problem) location field is missing from Achaea's response.") return end
```
vadi2: This has the potential to spam the user with a lot of error messages they wouldn't understand (and in this case, able to deal with), not really part of the current philosophy. Best to use a debugf here, so if something is wrong, you can look in the debug output and figure it out.

keneanung:
Right. I also replaced the echof at the place where I got... inspired.

vadi2:
+1

PR #90, opened 2015-10-17
fix dragon form detection

Seifur:
Colors were added to the value of gmcp.Char.Status.race. This changes it to look for Dragon instead of checking if it's equal to Dragon.


vadi2:
+1

PR #92, opened 2015-10-17
Fix #91

### LINKED ISSUE #91: table.deepcopy doesn't exist in Mudlet 2.1
There's an issue with commit https://github.com/svof/svof/commit/eff030c52a12836bbe009f74e090dc1db08fce97. table.deepcopy doesn't exist in Mudlet 2.1. These connected functions will need to be moved further down so they can use the already defined local function deepcopy.

keneanung:


Seifur:
Didn't see any typos and the spacing looks okay. You're just re-ordering the functions so that deepcopy is defined before the connect functions are added.

PR #98, opened 2015-10-29
Multiclass

### LINKED ISSUE #81: Multiclass Support
Now that multiclass is here, we need to be able to support it. Basically, this is how I see it working:

The system must be able to detect the current class the person is using.
The system must only use abilities granted for each class.
The system must be able to switch the defence screen to the currently used class.
The system must be able to remember user-defined defence lists for each class. (Preferably, the same alias (i.e. VDEFS BASIC) can be used for different classes and put up different defences. This could be done by keeping the general defence list the same between classes, and class specific defences stored with each class.)
The system must reset defences when the class is changed.
The system must recognize class change and adapt to it.

Most of all, this means that we're going to end up with only -one- release, one system to rule them all, that will work no matter what class the person is using. I have no idea how this will affect the final size of the system (though hopefully size will be reduced when we move afflictions and defences to gmcp).

This isn't an exhaustive list, and anything else that anyone can think of should be posted here so we can make sure we're working on it.

keneanung:
This pull request should start discussion on the implementation of multiclass as well as having a checklist so nothing gets forgotten.

This will fix #81 when merged.
- [x] Change `svo.me.class` on class change
- [ ] give an alternative to `svo.me.class` in the autoupdate code
- [x] stop defenses from overwriting each other in `raw-svo.defs.lua`
- [x] stop defenses from overwriting each other in `raw-svo.dict.lua`
- [x] filter available defenses by current class
- [x] defup/keepup only skills available in current class
- [x] go through `#if` preprocessor directives about class/skill specific logic
- [x] create and raise "classchanged" signals
- [x] remove most changes in `generate.lua`
- [x] ~~create a function from the systemstart signal handler that creates temp triggers and call it on class change~~
- [x] find a solution for TODO parts in `raw-svo.dict.lua`
- [x] fix up lifevision and (presumely) shroud. They get nixed during class change


raw-svo.dict.lua
```diff
@@ -14876,6 +14876,10 @@ signals.gmcpcharstatus:connect(function ()
     defences.got("dragonform")
   else
     defences.lost("dragonform")
+    if gmcp.Char.Status.class ~= me.class then
+      me.class = gmcp.Char.Status.class
+-- TODO handle other class changes here
```
vadi2: This would be a good place to emit a 'class changed' signal.

generate.lua
```diff
@@ -33,7 +33,8 @@ local args = lapp [[
 ]]
 
 local builder         = "lua" -- or "luajit"
-local doall           = args.release ~= "false" -- make doall a bool
+local doall           = args.release ~= "false" and args.name == "none" -- make doall a bool
+print("doall is " .. (doall and "true" or "false"))
```
Seifur: This could be `print("doall is " .. tostring(doall))` instead.

keneanung: Right. But most changes in generate.lua will be undone again. Those were some first tests with individual class cominations, but I believe those are impractical.

keneanung:
I updated the todo list in the first comment as well

keneanung:
@vadi2 Defenses should work now. I'd like to release an alpha preview of the allclasses packages tomorrow (with a big disclaimer). Any thoughts?

vadi2:
In my experience alphas don't work - people have a lot of miscommunication. The message will get simplified to "multiclass works", people will find out that it doesn't quite yet, and get disappointed. I'd recommend pushing out multiclass once you've got a lot of confidence in saying that it does work.

keneanung:
Hrm, all right.

caJaeger:
I noticed there is a task left to watch class status, this is what I have implemented:

Listener on gmpc.Char.status
```
function observe_char_status()
  local currentClass  = classManager.currentClass()
  local previousClass = classManager.previousClass
  if currentClass ~= previousClass then
    echo("class switch detected")
    classManager.mapIntegration()
  end
  classManager.previousClass = currentClass
end
```

and the class Manager:
```
classManager.currentClass = function()
  local class = gmcp.Char.Status.class:lower()
  if string.match(class, "dragon") then
    return "dragon"
  else
    return class
  end
end

classManager.mapIntegration = function()
  for class, skills in pairs(classManager.classSkills) do
    disableTrigger(class)
    for i, skill in ipairs(skills) do
      disableAlias(skill)    
    end
  end 

  enableTrigger(classManager.currentClass())
  for _, skill in ipairs(classManager.classSkills[classManager.currentClass()]) do
    enableAlias(skill)
  end
end
```

caJaeger:
What's the story here?

keneanung:
This is still a WIP, but I put some other things first for the upcoming release.

keneanung:
There are still some minor issues left to fix (auto update code among them), but otherwise, I believe this is ready for very thorough testing by a lot of people...

vadi2:
Good work!

keneanung:
@vadi2 and anybody else, who has an opinion about priorities, I need some input.

Due to the way I coded the class change for the dict table, priorities have to be reassigned after each class change. This by itself is not troublesome, however, new class specific things get assigned to lowest priorities because they were not included into the existing prios. That means the user would have to shuffle priorities manually after each class change. To solve this, I can imagine 2 solutions, which are both not ideal:

1. Keep a complete list of all possible priorities and operate on this list, reassigning priorities in dict on every change. This may be a relatively massive code change with the danger that the "all prios" list and the dict go out of sync. Also creating a sensible structure may be hard.
2. Create own priority lists for every class and load them on change. This may be the easier solution codewise. However, priority changes for one class are not reflected in other classes (which may be good, but is more often bad).

I don't know how to handle this and would like some input and thoughts.

vadi2:
Still having a think about this.

keneanung:
I started working on the first option. It should make external scripting easier if all priorities stay the same for all classes. Just imagine some important defenses jumping around depending on the number of defenses a class has...

caJaeger:
Does this branch wrap up all classes into one file or is it necessary to have the zip of each class?

As for your question, a bit late to the discussion, but I think there should be a dictionary that is class specific that loads on switch. Just my two cents. I would like to test this but I am unsure the status currently, looks like not much has been happening in the last few months. Is master waiting to be merged in?

keneanung:
It will create an additional package "Allclasses" with all classes in it. Specialized packages stay for now as they are often times more optimized.

Currently, the branch is pretty stale as life happened and I don't get much coding time.

adayoung:
Oops, it doesn't merge with the latest release. What is svo.me.class used for and can we not use `#if skills.<whatever>` instead of `#if class =`? :open_mouth: I want to help!

keneanung:
The `#if` are preprocessor directives. Those don't necessarily work with multiclass because we need to reconfigure available abilities and how they work during runtime. `svo.me.class` reflects the current class, so it can be used at runtime instead of `#if class`. Compile time changes are useful though, because of optimzation.

vadi2:
@keneanung let's close this since the in-client one has multiclass built in?

QwindorRousseau:
This has been migrated to #329

PR #111, opened 2015-12-01
Fix bindall and fortifyall abilities for sylvan

keneanung:
When merged, this will fix #75 and fix #47 

@slester can you confirm the fix again? I had Sarya test, but I'd like to have a second person look at it. A package is at https://dl.dropboxusercontent.com/u/7146752/Sylvan.Svof.v11.zip

@vadi2 I'd also appreciate a code review.

PR #113, opened 2015-12-03
Dont defer do dofree with batching

### LINKED ISSUE #79: Do/Dofree/Dor commands delayed after addbalanceful with batching
When using batching, the sending of dofree/do/dor (via expandAlias) is delayed until after the current batch, See https://github.com/svof/svof/blob/master/raw-svo.skeleton.lua#L741

If then things are added to the batch via sendc in svo.balanceful() to use up balance and thus cancel out dofree among other things.

keneanung:
This will fix #79 when merged. While usages of sendc with a table will break the current batch up, this makes sure we can run things we have no control over the way send is called in the queue (as it's done with expandAlias).

I was also stupid and included the commit of #111 here, so the merge should happen afterwards.

### DIFF
--- a/raw-svo.skeleton.lua
+++ b/raw-svo.skeleton.lua
@@ -733,15 +733,14 @@ make_gnomes_work_async = function()
   signals.sysdatasendrequest:block(cnrl.processusercommand)
 
   if conf.commandecho and (conf.commandechotype == "fancy" or conf.commandechotype == "fancynewline") then
+    oldsend = send
     send = fancysend
 
     -- delay expandAlias execution, used in dor, to after batch completes
     local oldexpandAlias = expandAlias
     if conf.batch then
       expandAlias = function(command, show)
-        tempTimer(0, function()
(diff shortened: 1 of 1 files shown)


vadimnehta:
You can submit a clean PR by making another branch and cherry-picking the commit into it.

raw-svo.skeleton.lua
```diff
@@ -733,15 +733,14 @@ make_gnomes_work_async = function()
   signals.sysdatasendrequest:block(cnrl.processusercommand)
 
   if conf.commandecho and (conf.commandechotype == "fancy" or conf.commandechotype == "fancynewline") then
+    oldsend = send
     send = fancysend
 
     -- delay expandAlias execution, used in dor, to after batch completes
```
vadimnehta: I would say the comment needs to be updated?

keneanung:
There we go. I used `git reset --hard` to reset to the commit before the sylvan changes and then cherry-picked the commit by sha and force pushed it.

vadi2:
+1 if you've tested this lots

PR #116, opened 2015-12-21
Fix high CPU usage after enchanting

keneanung:
oldsend is sent on startup, resetting it here is bad! Obviously a lot of testing is not enough :(


vadi2:
+1

Yeah, between batching and nice-looking send()'s, things get complicated. Might be worthwhile to draw up a diagram of the process.

PR #119, opened 2016-01-04
Change affliction of propagated Lobelia to Vertigo

keneanung:
With the classleads of 2016-01-02 thornrends with Lobelia propagated afflict
with vertigo instead of loneliness


vadi2:
+1

PR #121, opened 2016-01-04
Add the new shivering lines for hailstorm

keneanung:
I added the gain line in the "misc afflictions" part because those lines are actually generic and I have seen them in other places as well.

Or would it support anti-illusion if the attack line before was triggered as well?


keneanung:
@vadi2 any comments?

vadi2:
This setup will work I think.

PR #123, opened 2016-01-08
Title case class on generating

keneanung:
This should fix auto update issues.


keneanung:
@vadi2 Any comments?

vadi2:
+1

PR #128, opened 2016-01-18
Add reflection to weatherweaving

### LINKED ISSUE #127: Add reflection as def for sylvan
It seems like reflection was forgotten when sylvan got weatherweaving. See https://github.com/svof/svof/blob/master/raw-svo.defs.lua#L106

keneanung:
This will fix #127 when merged.


vadi2:
+1

PR #129, opened 2016-01-18
Add breathing defense line

keneanung:


raw-svo.defs.lua
```diff
@@ -210,7 +210,10 @@ defs_data = pl.OrderedMap {}
       end
 
       return true
-    end}) -- added in xml w/ conf.gagbreath
+    end,
+    def = [[You are holding your breath.]],
+    ondef = function() return "" end, --workaround to make it show up in defense lists...
```
vadi2: ?

keneanung: Yeah, this is a strange one.

Upon first analysis, I thought this was the only way to mark a defense with an X if it's not in the defc table.

After another inspection I'm not even sure how I traced it down to this... :cry:

vadi2: Without having tried this - I think all that would do it make it not show the defence name in the defences list. Do things work if you take it out?

keneanung: It will gag the line, but not tick the breathing defense...
The returned string gets saved in def_def_list and that gets appended to the def name (see https://github.com/svof/svof/blob/master/raw-svo.defs.lua#L3134)

vadi2:
+1

keneanung:
A little further inspection: Breathing is removed at the end of the def lines... So the flow is like this: def -> breathing def line shows -> breating is removed.

It's a strange thing to handle and not even consistently triggering the ticked box. I will need to have a closer look.

vadi2:
I think it'll be fine to have an FAQ on this explaining that by checking DEF, you're removing the defence, and not getting it back until the next prompt. If you try and fake it, logs will be confusing as someone can check def -> see that it's "there", get chocked or etc and take full damage, and wonder why did that happen.

keneanung:
Closing this as it's been here for a while... Maybe I'll open a new PR when it's ready.

PR #136, opened 2016-01-26
Added a missing line for clumsiness symptom

johndahlstrom:
I got this line as a Runewarden and noticed that it was nowhere to be found in the system.

### DIFF
--- a/svo (install the zip, not me).xml
+++ b/svo (install the zip, not me).xml
@@ -15251,9 +15251,11 @@ svo.valid.remove_unknownany(&quot;clumsiness&quot;)</script>
                         <colorTriggerBgColor>#000000</colorTriggerBgColor>
                         <regexCodeList>
                             <string>You lash out clumsily.</string>
+                            <string>^You lash out at (\w+) with (.+), but miss\.$</string>
                         </regexCodeList>
                         <regexCodePropertyList>
                             <integer>3</integer>
+                            <integer>1</integer>
                         </regexCodePropertyList>
                     </Trigger>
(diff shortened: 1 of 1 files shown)


svo (install the zip, not me).xml
```diff
@@ -15251,9 +15251,11 @@ svo.valid.remove_unknownany(&quot;clumsiness&quot;)</script>
                         <colorTriggerBgColor>#000000</colorTriggerBgColor>
                         <regexCodeList>
                             <string>You lash out clumsily.</string>
+                            <string>^You lash out at (\w+) with (.+), but miss\.$</string>
```
keneanung: Can clumsiness also make you miss denizens? Then you'd need to use `.+` instead of `\w+`.

johndahlstrom: After 5 minutes of bashing with clumsiness on me I did not see a single miss. So I'm guessing it's not making me miss denizens.

vadi2: +1

EldFitheach: I don't know how much we care about regex performance, in general, but it's probably better in principle to use lazy quantifiers with the more flexible wildcards, like "dot". That is, ".+?" instead of ".+", in this case. Also no need to save the match groups here, but no particular problem with having them there, either.

EldFitheach: A quick grep on some of my old logs turns up plenty of examples of this line applied to denizens, but they generally seem to just be missing due to poor accuracy (low Chivalry skill, to-hit, etc), rather than clumsiness. I don't know if clumsiness also gives this line, or just reduces your accuracy so that you sometimes miss when you otherwise wouldn't. Either way, this line doesn't seem to be a reliable indicator for the affliction.

keneanung: @EldFitheach good points all around. This seems to be not reliable enough as an indicator to rely on. @johndahlstrom @vadi2 any other thoughts? Otherwise I'd close this without merging.

vadi2: Yeah agreed, probably why it wasn't in before.

keneanung:
Closing due to discussion

PR #137, opened 2016-01-28
Fix possible nil reference to gmcp.Char.Status.race

EldFitheach:


keneanung:
:+1: looks simple enough. I assumed this kind of check would be necessary, but needed someone to confirm that there was no further issue.

vadi2:
+1

keneanung:
Another thought... Does dragonform stay through log out?

If so, maybe we should skip the whole function if we find no race field, because otherwise we'd first set "no dragon", which sets some things in motion and undo that decision afterwards.

vadi2:
Yes it stays through logout.

EldFitheach:
Yeah, that just occurred to me over my morning coffee. Fixing now.

keneanung:
Renewing my :+1: :smile:

QwindorRousseau:
Is this whole issue dealing with the dragon portion of
gmcp.Char.Status.race?

keneanung:
Um... No? If you read this and the issue it's about an error message that comes up on login and sometimes in between...

QwindorRousseau:
Okie dokie wasnt sure as i couldnt see it all in the email i got.

PR #141, opened 2016-01-30
Add palmstrike triggers

EldFitheach:
Not sure if I'm doing this right. Is there a problem with having the changes from another PR also registering in this one? Also sorry for the messy commit history - still figuring out git.


vadi2:
Look at squashing all the commits - that'll fix the messy history.

keneanung:
And possobly also at rebasing onto master.

Vadim Peretokin <notifications@github.com> schrieb am Sa., 30. Jan. 2016
10:57:

EldFitheach:
One thing I wasn't sure about with this was whether/how to include any anti-illusion checks, particularly for impatience. Should I be using svo.valid.proper_impatience() instead of svo.valid.simpleimpatience(), and/or is there anything else that should be done along those lines?

vadi2:
In Svo you should always use the proper_ if one is available. That gets murky though with the advent of serverside which is simply immune to illusions. I'll leave it up to @keneanung to decide if it's worth continuing with built-in AI.

EldFitheach:
Well, whether we continue developing it, there's probably no reason not to use what already exists. I'm not entirely sure about how proper_impatience() would work in this case, though. We've got a unique impatience line ("The palmstrike catches you in the temple and your head swims.") that always immediately follows the palmstrike line, so it looks to me like if I just call proper_impatience with no changes, the AI will ignore it for not being the first line. I could also add the palmstrike line to the list of previousline checks, but the trigger, as written, will only fire if we get both lines together anyway, so it's not clear that that would actually change the behavior.

Is there a reason to prefer the previousline check in proper_impatience over the trigger chain approach?

vadi2:
Add the line to previousline as well, yes.

Even if you use a triggerchain, paragraph_length will still be 2 so the function won't go further.

eewallace:
On Sun, Jan 31, 2016 at 12:38 AM, Vadim Peretokin <notifications@github.com>
wrote:

> Add the line to previousline as well, yes.
>
> Even if you use a triggerchain, paragraph_length will still be 2 so the
> function won't go further.
>
Right, I'm just trying to figure out whether using proper_impatience() with
the palmstrike line added to previousline will actually produce any
different behavior than using simpleimpatience() in the trigger chain. They
look functionally equivalent to me, but I'm still learning the ins and outs
of the system.

Issue #144, opened 2016-02-16
Sword and Shield affliction issues


keneanung:
> Sent By: Aerek On 2/16/7:45
> //So I know you're up to your ears right now, but currently, When a S&B Knight hits with Smash High, and you focus away one of the mental affs, Svo removes -all- mental affs. Extremely game-breaking once an S&B'er figures it out.

keneanung:
See https://ada-young.appspot.com/pastebin/42c19e6c

EldFitheach:
Are we sure the issue is for all mental affs, and not just recklessness? Would also be useful to know whether it happens with anti-illusion turned off.

Esync7:
Definitely all mental affs from the S&B attacks, and with anti-illusion off.

https://ada-young.appspot.com/pastebin/fb52b4f4

Esync7:
Better example from an actual fight. (From Nylian)

http://gotmanacards.com/combat_logs/mizik_snb_lock.html

Esync7:
Okay, this is not just an S&B thing. This is a global Svo issue. Svo clears all mental affs any time you focus, against any source.

https://ada-young.appspot.com/pastebin/7fa4ac50

vadi2:
Adding 25 bounty credits to this.

keneanung:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

Bounty received.
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBCAAGBQJW4uw9AAoJEE34WF2oeolCwq8IAJY5obbO/a6OrbsXTFReKw9K
YfnCfjf/r6ShObZZLK/MjbctTO31rSR+J7o6USf/1NzzPRFD3Yfs+Sv5FicotVs8
NWx9YpbNC6K0bWu6GLRx6TULVc88Y9Ph4e6d4B5khXvCdCfxjkTd2IUNSiqpv8g5
4mKiyh0OhcQd5+N7qbxe79QNEVEhDy3AGkrrecdjz/0derPstr9lII0cLnXn/LZ4
SMTDc7/BrCCfFR7pCAT/0CjHQxfexnwhtjQZsGpOSoXbdeu8ibjbWyjQ2/BsLeX5
aiYke6VhHFzWUCs9yuG/bQ3NsEGwS0hkIiRr5/MzfjPpxZqlvNye4IX1U3bTHws=
=m1IJ
-----END PGP SIGNATURE-----

PR #151, opened 2016-03-10
Remove weariness from focus_cured_* functions

keneanung:
This fixes svof/svof#144. It caused the "checkany" function to error out and
thus clear all focusable afflictions.


vadi2:
+1

Issue #150, opened 2016-03-08
Missing lemming passive line


mattnorr:
Sentinel's WOODLORE entity 'Lemming' passively afflicts with confusion, which svof does not have the line for.:
The lemming races around you and between your legs, confusing you with its eratic movement.
exact match ^
body:
svo.valid.simpleconfusion()

keneanung:
While somebody looks at lemming, here's another one 8reported by Deladan):
> ^A small brown lemming rips apart your (.+) with its claws.$

Should call `svo.valid.defstrip(matches[2])`

Ditsypop:
Went ahead and threw up a commit for this, doing 1 more thing then will do PR

Also @keneanung the trigger you referenced was already there when I looked.

keneanung:
I'd prefer single PRs for every issue, to be honest. Easier to track for me.

Ditsypop:
You got it boss. Filed.

keneanung:
About the trigger being already there: Maybe the message changed slightly? Can you test whether the defstrip is recognized?

Ditsypop:
taking a closer look and testing it out a few times to see what's up.

Ditsypop:
Updated!

PR #178, opened 2016-04-14
Lemming confusion added (issue #150)

Ditsypop:
Added trigger for lemming confusion, updated line for lemming def strip. as per issue #150

### DIFF
--- a/svo (install the zip, not me).xml
+++ b/svo (install the zip, not me).xml
@@ -48516,6 +48516,26 @@ svo.valid.simpleillness()</script>
                         <integer>3</integer>
                     </regexCodePropertyList>
                 </Trigger>
+                <Trigger isActive="yes" isFolder="no" isTempTrigger="no" isMultiline="no" isPerlSlashGOption="no" isColorizerTrigger="no" isFilterTrigger="no" isSoundTrigger="no" isColorTrigger="no" isColorTriggerFg="no" isColorTriggerBg="no">
+                    <name>Lemming confusion</name>
+                    <script>svo.valid.simpleconfusion()</script>
+                    <triggerType>0</triggerType>
+                    <conditonLineDelta>0</conditonLineDelta>
+                    <mStayOpen>0</mStayOpen>
+                    <mCommand></mCommand>
+                    <packageName></packageName>
+                    <mFgColor>#ff0000</mFgColor>
+                    <mBgColor>#ffff00</mBgColor>
+                    <mSoundFile></mSoundFile>
+                    <colorTriggerFgColor>#000000</colorTriggerFgColor>
+                    <colorTriggerBgColor>#000000</colorTriggerBgColor>
+                    <regexCodeList>
+                        <string>The lemming races around you and between your legs, confusing you with its eratic movement.</string>
+                    </regexCodeList>
+                    <regexCodePropertyList>
+                        <integer>3</integer>
+                    </regexCodePropertyList>
+                </Trigger>
             </TriggerGroup>
             <TriggerGroup isActive="yes" isFolder="yes" isTempTrigger="no" isMultiline="no" isPerlSlashGOption="no" isColorizerTrigger="no" isFilterTrigger="no" isSoundTrigger="no" isColorTrigger="no" isColorTriggerFg="no" isColorTriggerBg="no">
                 <name>Weaponmastery</name>


keneanung:
@vadi2 @EldFitheach any thoughts about the performance implications of removing the `Begin of line substring` guard? I know this is an ongoing discussion whether it's worth it or not, but I'm still not sure which policy to adopt here actually.

vadi2:
You certainly don't improve the performance by removing it, substring matching is going to be faster than regex matching any time. It's been a while since I've done the performance tests, but from what I recall on a simple pattern, regex was 1.5x slower.

keneanung:
Even though there are no wildcards or matching classes?

@Ditsypop Could you then keep the 2 matches and multimatches approach?

Ditsypop:
Of course, I misinterpreted the message and thought you meant you wanted the trigger to be exactly what you stated.

Ditsypop:
actually I do have one question, how do you want me to file it? On my own projects I always just... Overwrite the changes in another PR, which is probably not the best way to go about it.

keneanung:
It's best to create a new branch for every new feature/change. That way you can work on multiple things in parallel and add new commits to specific PRs.

Ditsypop:
Okay you got it boss will take care of it!

Ditsypop:
Hi I'm a nuisance one more question, did you want Lemming confusion as a substring even though it's always the same line? [image](http://i.imgur.com/NuLOooL.png)

keneanung:
No, that should be an `exact match`. The `Begin of Line` is a kind of guard because even though there are no wildcards, regex are a lot slower than begin of lines. The speed improvements are high enough to offset the penalty of running 2 triggers on the actual match.

Ditsypop:
Ehhhhh I resubmitted it and it went to an entirely different PR I feel a little confused, not familiar with git that much, sorry for all this inconvenience over something so ridiculously simple, I'm learning!

keneanung:
You can see what happened by looking at the [network graph](https://github.com/svof/svof/network): You created a new branch (wich is great) and created a new PR for it. Since the other 2 commits before that are not part of the main repository yet, they are part of the second PR as well.

As an explanantion: Git doesn't really know of objects called "branches". They are simply some kind of label or pointer that are pretty easily moved. And a merge tries to reconcile 2 pathes that have diverged so that the merge target includes all commits in the end.

keneanung:
The following sequence of commands on the command line should clean the PRs up:
~~~
git checkout master
git reset --hard fef4f3c47d4cc2bd9d574c1f5af7d9b7f17c00a6
git push --force

git checkout Def-strip-(lemming)
git rebase -i 3aa3aaf00d53887648a5380f3a078aeaece88508
#delete the first line in the editor and save
git push --force
~~~
The first block will isolate the confusion part and will put it into this PR.
The second block will isolate the def strip part and put it into #179 

Both times `--force` is needed, because you're changing history.

Ditsypop:
ok im trying that right now !

bunster:
Using git via the command line is a pain in the arse. I'd highly recommend using something like [GitKraken](https://www.gitkraken.com/) instead.

keneanung:
Each to their own. Depwnding on environment I actually prefer using the command line and since I didn't want to assume anything about the environment  I chose to give command line instructions, which are part of almost every tool.

Giving commands is also a lot easier than explaining GUI actions.

Ditsypop:
well i tried the command line thing but I got that error i linked and I'm not sure what it means.

keneanung:
I don't see a lin :(

bunster:
@keneanung GitKraken is multi-platform, so there's no need to worry about the environment. I don't know why you'd choose (let alone _prefer_) to spend time typing in six different commands on the terminal when it can all be done in one click... :pensive:

Ditsypop:
the link was just the word 'this' that linked to an imgur link [http://i.imgur.com/G9h80Lc.png](http://i.imgur.com/G9h80Lc.png)

vadi2:
SmartGit is a pretty nice client that I use.

keneanung:
I use GitKraken at work as well. But sometimes you are on systems which don't have that installed (only something else) or don't even have a GUI :fearful:

keneanung:
Yeah, you'll need to escape the parens (probably use \( and \)

Ditsypop:
I'm not sure if your message got formatted funny or something but I just see plain parenthesis things where you wrote to me.

keneanung:
Uh right. Use a backslash in front of both

Ditsypop:
Got it, I googled what the escape characters were in powershell which is I guess what I'm using and it's a little back-tic (`) mark. However after following your directions I'm still just getting heaps of error, I don't think I'm smart enough to handle this myself.

EldFitheach:
What errors are you getting once you add the backticks to escape the parens?

Ditsypop:
I dunno what to call it so I'm going to upload the entire contents of my buffer I guess. [https://spit.mixtape.moe/view/f6938fd6](https://spit.mixtape.moe/view/f6938fd6)

EldFitheach:
Sounds like there's probably a merge conflict (two commits that made different changes to the same line, so git doesn't know which version to use). If you still have that shell open and haven't done anything more since what you pasted, you can probably fix it by editing the file you made changes to (looks like that's the main svof xml file)..

To do that, you'd want to open the file in a text editor, and look for a section that looks like

    <<<<<< branch-name
         some code
    ======
         more code
    >>>>>> another-branch-name

The two sections of code would be two different versions of the relevant section. If you find a section like that, figure out which version is the one you want to commit, keep it, and delete the rest (the incorrect version as well as the <<<, ===, >>> lines). Then save the file, and back in the shell, do

    git rebase --continue

With a little luck, that should finish up without errors, and then you can go ahead with the push.

Ditsypop:
ugh okay I think I did it right now? This is very confusing I'm not very comfortable with command line stuff.

keneanung:
Almost, but the PRs are now at least separated :)

I'll merge them in the right order now.

Issue #164, opened 2016-03-22
Alchemist displace dismounts


keneanung:
> Sent By: Jukilian On 3/17/9:20
> // Hey. Seems like Svo doesn't pick up on you getting dismounted as the result of getting displaced by an Alchemist. I'm not sure if this is an Svo problem or a GMCP problem though.

This is an svo problem as long as svo doesn't use GMCP for defenses.

PR #196, opened 2016-06-10
Adding EtherDisplace to dismount lines

jgh713:
As per #164


keneanung:
Do we need the capture of the displacing person?

jgh713:
Suppose it's not necessary, by any means, but not perceptibly detrimental either. *shrug*

keneanung:
Probably not, but to keep it in line with te others, I'd prefer it to be adapted. Also: fancy wording ;)

PR #198, opened 2016-06-18
Deathsight changes

jgh713:
Thoth's death changed deathsight lines. Shouldn't be an issue when gmcp affs/defs go in. but until then here's the changes.


jgh713:
Every five minutes someone new loses their goddamn mind because of this. Pull ASAP please.

Issue #44, opened 2015-08-26
Transition to gmcp-based affliction and defense detection.


Seifur:
Implementing this would significantly reduce the size and complexity of the project.

##### There are a few things to note:  
Many abilities do not share the same name as the ones set in `svo.affl` and `svo.defc`
###### Afflictions
- **Bug:** There is no Add event when you gain an additional stack of a stackable affliction.
- **Bug:** There is no Remove event when you cure a level greater than 1 (cure level 2 -> now have level 1) of a stackable affliction
- **Add/List:** The name item contains a number in parenthesis for stackable afflictions.
- **Add/List:** The name item for humours does not contain a number
- **Add:** Hidden afflictions do not generate an event
- **Remove:** Is a 1 item length indexed table. The name is always just the base name. No numbers in parenthesis

###### Defences
(To be added later - needs more testing)

vadi2:
Are all defences tracked, such as elementalism channels, angel and demon defences, harmonics, morphs, lyre, etc? I think there would be gotchas in defences as well.

Seifur:
I think for defences it's only those that have a line in `DEF[ENCES]`, but would need to test various artefacts, harmonics, vibrations, etc.

Seifur:
Just remembered that a decent number of the server names for both afflictions and defenses aren't the same as the ones set in `svo.affl` and `svo.defc`. We would need something to translate them over to maintain compatibility.

vadi2:
The translation is already there as part of the serverside curing system, see "gamename" in svo.dict.

QwindorRousseau:
Unknown afflictins via loki and blight ect ect could still be tracked via triggers for the system like it currently is then send a diag after a certain or manually update the affliction list via manual diag. Same has been true for torso breaks for some time so still a certain part of player input needed to sometimes keep up to date on curing. But with the gmcp table for afflictions it would help greatly in curing even in further class updates unless they add some completely new skill that does something different like fractures system and it would help with bashing as well in new areas.

QwindorRousseau:
Would be a great leap for the system if we can get it to track via gmcp thats for dang sure as i know some people have messed with wundersys to track gmcp afflictions. The conversion would just take the time to test each affliction then pull it via gmcp then have it reference a table to set svo.gotaff i think is the api but not sure off the top of my head right now based on the gmcp name.

slester:
Is anyone working on this already? I don't want to repeat work.

Seifur:
[...]

PR #209, opened 2016-07-15
GMCP Affliction and Defense Tracking

jgh713:
Bleepbloop.

As far as I could tell, I didn't break anything. I can't promise things will remain that way.

And as I wrote that sentence, I just realized a relatively sizeable fuck-up. Pushing fix soon.


jgh713:
There we go :angry: 

Anyway, yeah. Whatever.

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+signals.gmcpcharafflictionsadd:connect(function()
+  local thisaff = gmcp.Char.Afflictions.Add.name
+  if thisaff:sub(-4) == " (1)" then thisaff = thisaff:sub(1, -5) end
+  gaffl[thisaff] = true
+  if conf.gmcpaffechoes then echof("Gained aff %s", thisaff) end
+  if dict.sstosvoa[thisaff] then
+    addaff(dict[dict.sstosvoa[thisaff]])
+  end
+end)
+signals.gmcpcharafflictionsremove:connect(function()
```
keneanung: Could you add a new line here?

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+  local thisaff = gmcp.Char.Afflictions.Remove[1]
+  gaffl[thisaff] = nil
+  if conf.gmcpdefechoes then echof("Cured aff %s", thisaff) end
+  if dict.sstosvoa[thisaff] then
+    removeaff(dict.sstosvoa[thisaff])
+  end
+end)
```
keneanung: Another new line here, please.

raw-svo.setup.lua
jgh713: Between each signals?

jgh713: Each connect, that is

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+signals.gmcpcharafflictionslist:connect(function()
+  gaffl = {}
+  local preaffl = {}
+  for key, val in ipairs(affl) do preaffl[val] = true end
```
keneanung: where does this `affl` come from? Only thing I can find further above is `affs`.

raw-svo.setup.lua
keneanung: Yes, please

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+  thisdef = gmcp.Char.Defences.Remove[1]
+  gdefc[thisdef] = nil
+  if conf.gmcpdefechoes then echof("Lost def "..thisdef) end
+  if dict.sstosvod[thisdef] then
+    defs["lost_"..dict.sstosvod[thisdef]]()
+  end
+end)
+signals.gmcpchardefenceslist:connect(function()
+  gdefc = {}
+  local predefs = table.deepcopy(defc)
```
keneanung: careful about `deepcopy`. Mudlet 2.1 doesn't have that yet, but svof supplies a local `deepcopy` function.

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+  for index, val in ipairs(gmcp.Char.Defences.List) do
+    thisdef = val.name
+    gdefc[thisdef] = true
+    if dict.sstosvod[thisdef] then
+      if predefs[dict.sstosvod[thisdef]] then
+        predefs[dict.sstosvod[thisdef]] = false
+      elseif type(defs["got_"..dict.sstosvod[thisdef]]) == "function" then
```
keneanung: You are checking `got_def` for a function here, but you are not doing that in `gmcpchardefencesadd` above.

raw-svo.setup.lua
```diff
@@ -377,6 +471,12 @@ signals.healingskillchanged         = luanotify.signal.new()
 signals.morphskillchanged           = luanotify.signal.new()
 #end
 
+
+
+
+
+
+
```
keneanung: Why this empty block?

raw-svo.setup.lua
```diff
@@ -256,6 +259,97 @@ signals.gmcpiretimeupdate:connect(function()
...
+  for key, val in ipairs(affl) do preaffl[val] = true end
+  
+  for index, val in ipairs(gmcp.Char.Afflictions.List) do
+    local thisaff = val.name
+    if thisaff:sub(-4) == " (1)" then thisaff = thisaff:sub(1, -5) end  
+    gaffl[thisaff] = true
+    if preaffl[thisaff] then
+	  preaffl[thisaff] = false
+	elseif dict.sstosvoa[thisaff] then
+      addaff(dict[dict.sstosvoa[thisaff]])
```
keneanung: The indenting looks weird here.

raw-svo.setup.lua
jgh713: affl is svo's list of current afflictions. Wasn't actually sure how affs was structured, so just stuck with affl, as iterating over the index is quick and easy.

keneanung: Yeah, I found that one meanwhile as well. The woes of naming...

raw-svo.setup.lua
jgh713: It's an extra failsafe that is left over from an earlier version of the script. There should never be an instance in which sstosvod[thisdef] exists and there's no got_def function for the def. If there -is- such a situation, erroring in a way that we run into it would be better than just failing silently, so it might be best to remove this, or add a 'Hey tell someone this happened echo.'

raw-svo.setup.lua
jgh713: Not sure where this came from actually.

raw-svo.setup.lua
jgh713: affs is a local dictionary, I think. affl is globally accessible. Still not sure of the structure of affs, but I think it's more than just the basic list that affl is.

keneanung: Yeah. It's ok using this, I was just statically looking at this and wondering where that table came from ;)

raw-svo.setup.lua
jgh713: Up to you, really. I don't care. Probably either an else echof("Hey fuckface we broke") in place of the following end, or just replace that particular elseif with an else.

Or something else, I guess.

keneanung: Maybe we could add the usual error reporting of svof by adding the (e!) link to the begin of the line? But we'd need to add it to the add above as well.

raw-svo.setup.lua
```diff
@@ -256,6 +259,101 @@ signals.gmcpiretimeupdate:connect(function()
...
+  gdefc[thisdef] = nil
+  if conf.gmcpdefechoes then echof("Lost def "..thisdef) end
+  if dict.sstosvod[thisdef] then
+    defs["lost_"..dict.sstosvod[thisdef]]()
+  end
+end)
+
+signals.gmcpchardefenceslist:connect(function()
+  gdefc = {}
+  local predefs = table.deepcopy(defc)
```
keneanung: careful about `table.deepcopy`. Mudlet 2.1 doesn't have that yet, but svof supplies a local `deepcopy` function.

jgh713:
You're correct.

jgh713:
"The problem was: lost_ function was nil for defence abctest (gmcp: xyztest)"

keneanung:
It's correct for the `Remove` function. For the `Add` and `List` functions you are doing

    [[echo("The problem was: got_ function was ]]..type(defs["**lost**_"..dict.sstosvod[thisdef]])..[[ for defence ]]..dict.sstosvod[thisdef]..[[ (gmcp:]]..thisdef..[[)")]]

(embedding the type of the `lost_` function instead of the `got_` function)

jgh713:
Oooooooh, you're right.

Though, technically speaking, if one doesn't exist, neither should the other. If that's the case, that's an even bigger issue.

jgh713:
GDI

jgh713:
Anything else @keneanung? Things seem to work on my end, but some actual testing by someone other than me before pushing a change like this would be nice.

keneanung:
@jgh713 definitely agree. But right now I can't test this (I'm not active at all). Do you want me to create a package that someone can test?

jgh713:
If you want to, that'd be great. Or, you can merge and just be ready to pull bugfixes asap :>

keneanung:
Here you go: https://www.dropbox.com/sh/xmlznnnichmgqd5/AAAsH0kz7u3d1Q6GpxGo6NrWa?dl=0

jgh713:
Unsurprisingly, svo exploded. Will figure out why when I'm less sleep-deprived.

keneanung:
related issue: #44

keneanung:
Hrm... I'm not sure why this fixes the issues you were having. I can't see any broken closures or anything.

keneanung:
Another point I'm just noticing: wouldn#t you want to check, if server side curing is enabled, before sending a predict?

jgh713:
@keneanung maybe? I mean, there's not really a downside to predicting regardless, and it sets them up in the rare cases where they're swapping? Idk.

Will add more explosion fixes later :s

keneanung:
I'd just be confused if I used pure svof (maybe for a reason?) and suddenly some `[CURING]` messages popped up.

jgh713:
Fair enough, though I think the line is just "You have added the AFFNAME to your afflictions." or something. *shrug*

I'll do that :+1:

PR #211, opened 2016-07-27
Battlerage

jgh713:


raw-svo.setup.lua
```diff
@@ -244,6 +244,12 @@ signals.gmcpcharitemsremove:connect(function()
   end
 end)
 signals.gmcpcharvitals      = luanotify.signal.new()
+signals.gmcpcharvitals:connect(function()
+  if not gmcp.Char.Vitals.charstats then return end
+  for index, val in ipairs(gmcp.Char.Vitals.charstats) do
```
vadi2: How does the structure of charstats look like? There's a possibility of stats.battlerage being set, and then reset on a subsequent iteration because the Rage: string isn't found.

jgh713:
@vadi2: val:match("^Rage: (%d+)$") or stats.battlerage or 0

At no point should that reset a previous value, unless I'm a complete dolt.

And it's just an indexed table of strings, e.g. { "Bleed: 0",  "Rage: 0",  "Venom: " }

vadi2:
allg, I misread it for something else.

jgh713:
Yay I'm not a dolt :p

PR #213, opened 2016-07-27
Improve travis to actually return exit code != 0 on failure

### LINKED ISSUE #212: Travis.ci build
So the travis build doesn't actually fail if there's an issue and it can't complete the build. It just errors on the first class package (shown in the log) then exits as a successful build. Not sure how relevant this is, but it'd be nice to not have to scrutinize the log each time to see if things break when pushing changes.

Idk, just seems logical.

keneanung:
fixes #212

### DIFF
--- a/generate.lua
+++ b/generate.lua
@@ -25,7 +25,7 @@ local args = lapp [[
 -d,--debug  Build with debugging information enabled
 -r,--release (default false)  Build all systems
 -o,--own  Build a Svof for yourself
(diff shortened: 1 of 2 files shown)


keneanung:
@jgh713, @vadi2 anyone of yiu feeling like critiquing this?

vadi2:
Looks alright. Does it work?

keneanung:
Just reaffirmed: It does :smile:

Issue #215, opened 2016-07-29
Cloaking and shroud are the same defence.


jgh713:
Should be merged into the same defence. Not sure what all that's going to entail, or if that's going to wreck anything, but they're the same, mechanically speaking. Shadowcloak, serpent cloaking and necromancy shroud give you the same defence, but are handled separately (shadowcloak is necro shroud).

jgh713:
Best way to do this is probably just change serpent cloaking to shroud.

Maybe add got_ and lost_cloaking functions as an alias for shroud, for compatability reasons.

jgh713:
Going to wait till someone else weighs in though, in case I'm missing some obvious explosion waiting to happen.

keneanung:
I can't really see any issues. So if you want go ahead, whoever wants to take this.

keneanung:
#246 is a workaround for the gmcp part of the issue but in the long run, they should be merged.

QwindorRousseau:
@svof/developers again take a look and see and do a quick fix for it. Will probably need to add in a class check to do the right command for it but should be fairly easy.

QwindorRousseau:
there isnt much of a reason to do this because of the different commands for each. the system recognizes it as the same defense in the end with the gmcp and the lines but need to keep commands seperate.

PR #216, opened 2016-07-29
Changing shroud line to regex

jgh713:
For customised shrouds

### DIFF
--- a/raw-svo.defs.lua
+++ b/raw-svo.defs.lua
@@ -407,7 +407,8 @@ defs_data = pl.OrderedMap {}
     def = "You are emanating an aura of death.",
     off = "Your aura of death has worn off."})
   defs_data:set("shroud", { type = "necromancy",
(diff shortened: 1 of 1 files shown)


keneanung:
Could this be obsolete with gmcp defs? My plan was to slowly fade def/aff triggers out now.

jgh713:
Will have to code in an exception that defs both shroud AND cloaking for this defup, or they'll need to be combined as per #215 (which should be done anyway, since they're the same), but essentially, yes.

jgh713:
Though, be careful about phasing out those triggers, there's a decent number of things that gmcp doesn't cover adequately.

keneanung:
I do. This is why I want to phase them out instead of deleting them right away.

jgh713:
No longer necessary with a recent change Ken made to dict.sstosvod. Going to close.

PR #218, opened 2016-08-07
A fix for the ndb alias cheatsheet erroring out

ndsamuelson:
I am uncertain WHY it was erroring every time, but in order to make this function this was the only solution I could find to the problem.

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -2647,21 +2647,6 @@ send(matches[2]..&quot; enemies&quot;, false)</script>
 if not ndb.help then
   local path = getMudletHomeDir()..&quot;/&quot;..class..&quot; svo&quot;
   if not lfs.attributes(path) then
-  -- not found? Lets try the dev version
-    local svo_location = getMudletHomeDir() .. _sep .. &quot;svo_location&quot;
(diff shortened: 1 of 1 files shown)


keneanung:
This will break the cheat sheet for developers, who don't install svof as a package.

ndsamuelson:
Well, I'm just not sure what is causing the script to return nil for f when that code is in there. It may break the cheatsheet for developers, but for the everyday user who needs it more I would say that it would better serve everyone if we could figure it out. I'll mess with it tomorrow when I have more time

keneanung:
It will use this part only if the normal location is not found. Maybe we should find out why this is the case.

Are you using a recent svof? Do you use the correct class package?

ndsamuelson:
I am using the most updated version. My mudlet home dir just for your info is... C:\Users\Nic_2\.config\mudlet\profiles\Colg clean\

I'm messing with it right now to see if I can solve it. As it stands right now it's not working again. So I went back to the original code to try and find a reason f returns nil on line 49.

ndsamuelson:
Going to make a correct pull request. Sorry for the confusion. I'm new here :)

PR #219, opened 2016-08-08
A (hopefully) viable way to fix the ndb alias

ndsamuelson:
Added a way to check which route the script took. I tested on my end and it seems to work. I don't run a dev version though so I can't test that side.

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -2677,10 +2677,10 @@ if not ndb.help then
   if not system_path then 
 	fullpath = path ..&quot;/ndb-help.lua&quot; 
 	local f, m = io.open(fullpath)
-	local s = f:read(&quot;*a&quot;)
+	s = f:read(&quot;*a&quot;)
   else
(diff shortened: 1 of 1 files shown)


ndsamuelson:
And I have no idea why travis-ci is erroring out. Or really what travis-ci is even. Heh.

keneanung:
Travis is a tool which runs our (minimalist) integration tests. Them failing isn't your fault though (in this case). Looks like a tool we use updated its dependencies and one of those uses a ruby version that is not installed by default.

ndsamuelson:
Well that's good. Glad it wasn't my fail coding heh

ndsamuelson:
I have confirmed that this fixes the problem at least for those not using a dev version of svof. Can someone try it on the dev version to make sure it works properly?

keneanung:
I opened an alternative PR (#224) with the actually preferred way to go. But this PR got me on the right track. Thank you very much :+1:

PR #220, opened 2016-08-08
Update .travis.yml to use a higher ruby version

keneanung:
looks like one dependency for the github_changelog_generator requires ruby
2.2.2 or higher, while travis only supplies 2.2.0 by default. This should
hopefully fix the error.


keneanung:
Merging this as probably nobody else wants to have a say and travis is green again :P

PR #223, opened 2016-08-13
Make sure howMany parameter of getSameItems is a number

keneanung:
Earlier we compared numbers to strings (`string.match` with `%w` returns a string) and thus never stopped returning items for enchanting.


jgh713:
Looks good :D

PR #224, opened 2016-08-13
Fix the ndb cheat sheet

keneanung:
Formerly `path` only contained the actual path, no cheat sheet file, which of course can't be opened as a file. This fixes this by checking the path inclusive file for existance.


ndsamuelson:
Well that is far more elegant than my solution hah

jgh713:
Me likey :sake:

PR #228, opened 2016-09-09
Add bubble pipe

keneanung:


jgh713:
Looks perfectly fine to me. Seems to work with regular pipes, so it's not -completely- broken. :+1:

PR #232, opened 2016-10-10
Fix deathsight line

keneanung:
This is necessary, even though it should not be. Deathsight is transmitted via
gmcp but doesn't work correctly... somehow...


jgh713:
This seemed to be an issue with nightsight at one point as well.

PR #236, opened 2016-11-03
Improvement: use GMCP instead of QL to request room content

keneanung:
This should save us some spam when being in a room without known Ourobori.


vadi2:
+1 if that works now!

keneanung:
@vadi2 care to formally approve a review?

keneanung:
I tested it and will do so again after I released the "testing" release.

Issue #238, opened 2016-11-05
New Class: Depthswalker


jgh713:
New class: Depthswalker

Skillsets: Aeonics, Shadowmancy, Terminus

New Defences:

  In Aeonics:
  
    Blur
	on: With a minor application of will, you increase your speed.
	off: This defence has no removal message, nor are gmcp events working properly for it. Seems to last 8:20 unboosted and 16:45 boosted.
	command: "chrono blur" or "chrono blur boost", preferably based on a 'vconfig boostblur' or some such. If no vconfig, then default no boost.
	
	
  In Shadowmancy:
	
	Shadowcloak
	"a grim cloak"
	"Shadow cloak" to summon.
	Requires full mana to create or summon if not present.
	
	Disperse
	gamename: disperse
	"Shadow disperse" to def.
	Requires shadowcloak, does not fade, takes equilibrium.
	
	Shadowveil
	gamename: shadowveil
	"Shadow veil" to def.
	Requires shadowcloak, does not fade, takes equilibrium. This ability also defs hiding.
	
  
  In Terminus:
  
    All of the following abilities require "word balance", which will need to be tracked via triggers, as there is no gmcp for it.
  
    Trusad
	gamename: precision
	command: intone trusad
	
	Tsuura
	gamename: durability
	command: intone tsuura
	
	Kail
	gamename: prismatic
	command: intone kail
	Requires and consumes eq, as well.
	
	Ukhia
	gamename: ?
	command: intone ukhia
	
	Qamad
	gamename: ?
	command: intone qamad
	
	Tah'maal
	No gamename, not tracked. Presumably a one-time empowerment, similar to runic armour. Likely not necessary to track, and overly complicated.
	
	Mainaas
	gamename: bodyaugment
	command: intone mainaas
	
	Gaiartha
	gamename: antiforce
	command: intone gaiartha
	
Active cure ability: 

    Accelerate
    command: chrono accelerate [boost]
    Boosted cures two random afflictions, otherwise it cures one. Requires and consumes eq
    use: You accelerate your body's ability to purge afflictions a thousandfold.
    off cd: You may accelerate time to cleanse your afflictions once again.
	
	
New tracking:

  Word balance:
  Is used for Terminus abiliites. 
  Balance returned: You may intone another word of power.
  Balance lost:
	Your mouth falls open as the all-consuming litany pours forth, and you know that in moments all that hear shall serve your will.
	^Imbuing your voice with power, you intone, "\S+"\.$
[...]

PR #239, opened 2016-11-05
Add depthswalker class package and defenses

keneanung:
Adds support for the new depthswalker class defenses.

See #238 

- [x] Add shadowcloak
- [x] add disperse
- [x] add shadowveil
- [x] add chrono blur
- [x] add word balance tracking
- [x] add age tracking
- [x] add Trusad
- [x] add Tsuura
- [x] add Ukhia
- [x] add Qamad
- [x] add Mainaas
- [x] add Gaiartha

Optional:
- [ ] Implement `med` equivalent for dilation
- [ ] add Kail


jgh713:
Age needs illusion/lag handling similar to other balances.

keneanung:
@jgh713 There you go. I wouldn't say age tracking is a balance, but a resource, which is why I implemented it slightly different.

keneanung:
@jgh713 I believe we are done for this circle, don't you think?

jgh713:
Looks like.

On Nov 22, 2016 5:01 PM, keneanung <notifications@github.com> wrote:

@jgh713<https://github.com/jgh713> I believe we are done for this circle, don't you think?

PR #243, opened 2016-11-11
Gmcp aff/def fixes

jgh713:


raw-svo.setup.lua
```diff
@@ -926,7 +822,7 @@ local index_map = pl.tablex.index_map
 local addaff, removeaff, checkanyaffs, updateaffcount
 
 local lostbal_focus, lostbal_herb, lostbal_salve, lostbal_purgative, lostbal_sip
-sk.salvetick, sk.herbtick, sk.focustick, sk.teatick, sk.purgativetick, sk.siptick, sk.mosstick, sk.dragonhealtick, sk.smoketick, sk.voicetick = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+sk.salvetick, sk.herbtick, sk.focustick, sk.teatick, sk.purgativetick, sk.siptick, sk.mosstick, sk.dragonhealtick, sk.smoketick, sk.voicetick, sk.wordtick = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
```
keneanung: This shouldn't be in this  PR, should it?

raw-svo.setup.lua
```diff
@@ -1019,6 +915,112 @@ local function assert(condition, msg, extra)
...
+  end
+end)
+
+
+signals.gmcpchardefencesadd:connect(function()
+  thisdef = gmcp.Char.Defences.Add.name
+  gdefc[thisdef] = true
+  if conf.gmcpdefechoes then echof("Gained def "..thisdef) end
+  if dict.sstosvod[thisdef] then
+    if type(defs["got_"..dict.sstosvod[thisdef]]) == "function" then
```
keneanung: Why are you checking the type of `dict.got_*` here if you call `defences.got()` below? The same goes for all other functions blow.

jgh713: @keneanung because the got_ functions are still generated normally, and are a reliable check for whether or not svo uses/recognizes the defence.

raw-svo.setup.lua
jgh713: And yeah, not sure how this snuck by.

raw-svo.setup.lua
keneanung: do we have defences is sstosvod, that are not in dict?

jgh713: Not currently, no. Unless you're a depthswalker using another package, in which case yes. It's just a warning system for when things go wrong, which hopefully they never should. Though it should probably be changed to warn you if a defense is nil in dict.sstosvod as well, so we're aware that there's a defense we don't have in the system. Which, as of DW changes, shouldn't be the case.

keneanung: It's just confusing that we check for something, never use it and use something entirely else instead... We either need a comment here or find a better way because the next person (or me in two weeks) stumbles on that part of the code and thinks something was forgotten to be changed like I did when I looked at it.

keneanung:
Was this closed in relation to #244?

Issue #202, opened 2016-07-06
Outdated Installation Instructions for Devs


caJaeger:
On OSX, the instructions for installing lua and the dependencies is outdated, brew simply doesn't know how to associate luarocks, penlight, and the filesystem to lua5.1. It will always refer to lua 5.2.

(I have before created an entire build with 5.2, why stick with a deprecated version of lua?)

Ignoring the brew instructions I have successfully installed the 5.1 version of lua, building it from source, as well as the most recent version of luarocks, also from source. All dependencies installed as well. Everything checks out, with the exception of the actual system, attempting to generate a class fails with this error:

```
generate.lua:181: attempt to index field 'skills' (a nil value)
```

I would love to start looking into this but every time I do I spend a few hours trying to get the basic environment setup and when I do, the code doesn't run.

Perhaps i am just missing something.

keneanung:
The build system may not need 5.1 (not 100% sure about that), but generally we stick to 5.1 because mudlet only supports that version.

I'll have a look into this issue later on, but in case it's a Mac specific issue, I can't help, cause I'm a linux user.

caJaeger:
@keneanung Alright, that's good to know. Perhaps I will just stick to building in 5.2 and make a VM with Ubuntu or something.

keneanung:
Just because I fiddled with Docker a lot lately and they released "native" Windows and Mac support (at least for some versions, but it worked okay on a windows 7 machine I tested), would you be willing to test that path for Mac?

It would allow Mac-Users to simply install docker and run a single command for building...

natecox:
I'm a big fan of docker, but I think it may be a bit much to install just for this.

Install instructions for macOS should specify `luarocks-5.1 install <package>` at the least. As an aside, I switched to [asdf](https://github.com/asdf-vm/asdf) to manage Lua so this wasn't an issue for me. Highly recommend.

QwindorRousseau:
This will no longer be a problem for developers as the bulk of the editing can be done in mudlet itself.

PR #245, opened 2016-11-21
Fixing osx install instructions

jgh713:
Fixing typo in osx install instructions.


jgh713:
Perhaps related to #202. Needs more work, because luarocks by default installs those packages to 5.2 install dirs, not 5.1

jgh713:
These instructions seem to work properly.

keneanung:
> Needs more work, because luarocks by default installs those packages to 5.2 install dirs, not 5.1

Is his fixed by installing a recent version of `luarocks`?

jgh713:
@keneanung luarocks searches for your current version of Lua, preferring 5.2 over 5.1. By not installing 5.2, it defaults to 5.1.

keneanung:
What if lua 5.2 is already installed for some reason? Maybe we should give instructions along http://stackoverflow.com/a/20359102?

jgh713:
That's why I added 'brew uninstall lua' to the instructions, but I suppose if they have it installed for another program or some such, that would work better.

jgh713:
Seemed to work fine on my system.

keneanung:
don't you have to change the luarocks command to `luarocks-5.1`?

jgh713:
I actually have no clue how the luarocks command works with different versions. It's just luarocks for 5.1 without 5.2 installed. Will have to look into it.

keneanung:
@jgh713 any updates on this?

keneanung:
All right, I dove a bit more into this. If you install `lua51`, luarocks will be installed automatically as `luarocks-5.1`, so no need to manually install luarocks-5.1, just change the luarocks command. A caveat: 5.2 and 5.1 rocks will overwrite each other. If that's a problem, the --tree option must be given.

vadi2:
No longer needed with upcoming in-client-svof.

PR #247, opened 2016-11-23
Adding bubble pipe to rebounding tracker

jgh713:


svo (reboundingsileristracker).xml
```diff
@@ -236,7 +236,7 @@ svo.rt_razed = nil</script>
                 <colorTriggerFgColor>#000000</colorTriggerFgColor>
                 <colorTriggerBgColor>#000000</colorTriggerBgColor>
                 <regexCodeList>
-                    <string>^\w+ takes a long drag off (?:his|her) ?.*pipe\.$</string>
+                    <string>^\w+ takes a long drag off (?:his|her)? ?.*pipe\.$|^A great gurgling sound is heard as \w+ takes a long drag from an enamel bubble pipe\. A series of shining bubbles erupt from the bowl of (?:her|his) pipe\.$</string>
```
vadi2: Have you tested this successfully? Mudlets regex does not support multiline, you'd have to make a multiline trigger here

jgh713: I have, yes. It's not a multiline trigger.

vadi2: Didn't notice the pipe character. Why not make another pattern in the same trigger though? It would be more readable and be less concerning on performance.

jgh713: What do you mean? It is in one trigger. You mean like:

` ^(?:\w+ takes a long drag off|A great gurgling sound is heard as \w+ takes a long drag from an enamel bubble pipe\. A series of shining bubbles erupt from the bowl of) (?:her|his) .*pipe\.$ `

that?

Or you mean disaable multiline trigger and set it in a second box? If that's what you're asking, it already is a multiline trigger and the second line's a relevant lua function.

keneanung:
@vadi2 @jgh713 any updates on this? Is this good to go?

PR #248, opened 2016-11-23
Depthswalker fixes

keneanung:
Fix things that showed up in the testing release (part 1 of ?)


jgh713:
@keneanung If someone was previously a serpent who had hiding on defup/keepup, that config remained active. As depthswalkers, they cannot enable or disable hiding, but svo still tries to defup/keepup hiding if they already had it. With this change, nothing seems to break and svo just completely ignores attempting to defup hiding, but still recognizes the defense as it would any other. However, It might be better to just change the definition of shadowmancy hiding so it's the same as shadow veil, so that hiding can be put on keepup as well. I'm unsure if there's any scenarios where you can lose hiding but not shadow veil (I know emerging loses you both).

keneanung:
Possibly one can be stripped independently of the other?

jgh713:
As a self note for later, or if you have time before that, shadowcloak still hangs if you lesserform while already wearing cloak, or if you're not wearing the cloak and svo has it marked as if you were and thinks you already have cloak. Need to add a simple trigger for the second, not sure what to do for the first.

zylvyn:
@keneanung If someone uses search, it removes hiding, but not shadowveil

keneanung:
@jgh713 I believe I fixed the shadowcloak issues, but I had no dragon around to test... I'll push the fix, maybe we can find someone.

keneanung:
@jgh713: @zylvyn tested the shadowcloak fixes. They seem to work!
About the hiding, I was designing hiding so you couldn't put it on keepup and defup, but I have no idea what I was thinking then... So maybe we should allow it on keepup, but probably autoenable shadowveil too in that case.

jgh713:
@keneanung Will test later tonight or tomorrow

jgh713:
Ah, cool. Should probably just leave them as two separate defenses. If people want to re-def on veil being stripped but not hiding, or vice versa, I dont see why not.

jgh713:
Where did you push the shadowcloak fixes? Dont see them.

keneanung:
Because I didn't push yet. Done now: ea523b6

keneanung:
@jgh713 I believe we are through with the issues, right?

jgh713:
Svo keeps up shadowcloak regardless of whether or not it's on defup/keepup (with no dependant skills.

PR #258, opened 2016-12-02
Add a Depthswalker specific lyre mode

### LINKED ISSUE #255: add lyre equivalent for Kail
Terminus Kail creates the same prismatic barrier like lyre. Incorporating it into lyre mode would be useful.

keneanung:
Similarily to druid and sentinel, Depthswalker brings an own lyre ability. I
added a way to use that for lyre mode.

Fixes #255 when merged

### DIFF
--- a/raw-svo.dict.lua
+++ b/raw-svo.dict.lua
@@ -15183,6 +15183,104 @@ affinity = {
       end
     }
   },
+  lyre = {
+    physical = {
+      aspriority = 0,
+      spriority = 0,
+      balanceful_act = true,
+      def = true,
+      undeffable = true,
+
+      isadvisable = function ()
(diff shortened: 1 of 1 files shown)


raw-svo.dict.lua
```diff
@@ -15183,6 +15183,104 @@ affinity = {
...
+
+      oncompleted = function ()
+        defences.got("lyre")
+
+        if conf.lyre then conf.paused = true; raiseEvent("svo config changed", "paused") end
+      end,
+
+      ontimeout = function()
+        if conf.paused and not defc.lyre then
+          echof("Lyre strum didn't happen - unpausing.")
```
keneanung: That'd be inconsistent with grove barrier and nightingale melody. Both talk about lyre, but are not 'lyres'. Also as soon as the depthswalker is dragon or uses a lyre as depthswalker by configuring the command, the echo doesn't fit either.

An additional point is that the aliases are "lyr", "lyc" and "lyf", which suggests lyre as well.

keneanung:
@vadi2 what do you think about my comments?

keneanung:
Same could be said about echoes and patterns: Once you know the pattern of an echo, same type of skill with the same echo help to recognise it in spam.

I'll merge the change in, but keep thinking about how to change the echoes.

vadi2:
Yeah, you raise a valid point, though the echo wouldn't change that much -
just the initial name. The rest of it would still be the same text

Issue #265, opened 2017-01-02
Svo not updating serverside focus settings


jgh713:
Two issues that are linked closely, and should be relatively simple to fix.

Vconfig focus doesn't configure serverside focusing as it should. Toggling this option should config the option accordingly, and svo should re-disable it if serverside focuses with the option disabled.

Svo needs to disable serverside focus when you are afflicted with cadmus (if conf.focuswithcadmus), and reenable it when cadmus is cured.

PR #266, opened 2017-01-06
Fix sk.canfocus function to return false if it shouldn't focus

keneanung:
While svof was prepared to toggle server side focus on cadmus, the can focus never returned false and thus never turned focus off.

This fixes svof/svof#265 when merged


keneanung:
This should be ready to go now as well.

PR #268, opened 2017-01-09
Gmcp anti-illusion failure workaround

jgh713:
Simple workaround for gmcp afflictions and defences failing due to anti-illusion checks.

There's likely a better way to do this already implemented in the system, but this works as-is.

### DIFF
--- a/raw-svo.defs.lua
+++ b/raw-svo.defs.lua
@@ -3021,12 +3021,12 @@ signals.systemstart:connect(function ()
         local bal
         for kk,vv in pairs(dict[k]) do if type(vv) == "table" and kk ~= "gone" then bal = kk break end end
         if bal then
-          defs["got_" .. k] = function ()
+          defs["got_" .. k] = function (force)
 #if skills.metamorphosis then
             if v.custom_def_type or usingbal(bal) then checkaction(dict[k][bal], true)
             else checkaction(dict[k][bal]) end
 #else
-            if not v.custom_def_type then checkaction(dict[k][bal]) else checkaction(dict[k][bal], true) end
+            if not v.custom_def_type then checkaction(dict[k][bal], force) else checkaction(dict[k][bal], true) end
 #end
             if actions[k .. "_" .. bal] then
               lifevision.add(actions[k .. "_" .. bal].p)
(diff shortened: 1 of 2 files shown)


keneanung:
I'm not aware of afflictions failing due to anti-illusion. Defenses failing is due to the call to "checkaction" in some got_* functions. My personal way to fix it would be https://github.com/svof/svof/compare/master...keneanung:gmcpAillusionFix

vadi2:
That would be a much better solution

jgh713:
Will need to add it to the other got_ function as well. Cant right now, but if it hasnt been done later I add it and test.

keneanung:
Which other `got_ ` function? The one that call only `defences.got`? Not really, because that has no illusion check at all.

jgh713:
Theres two calls for got_, one for gmcpchardefencesadd and one for list. I might have just missed the other one (on kobile right now).

keneanung:
Oh, right, forgot about the list!

keneanung:
Updated my branch for the list as well!

jgh713:
The fix doesn't work properly. There's some anti-illusion failsafe that I'm missing here. Even directly calling defences.got() fails because of aillusion for some reason. No clue why. @vadi2 any insight?

keneanung:
I have some insight. `defs.got_*` adds to the lifevision system, which waits for a prompt before processing. Between gmcp and prompt comes the curing command, which routes through to `killaction()`. This function also removes from the lifevision processing list and thus the gmcp update is lost.

keneanung:
Updated my branch (https://github.com/svof/svof/compare/master...keneanung:gmcpAillusionFix) with something that worked. @jgh713 you want to check it again?

jgh713:
I havent tested your changes yet, but directly calling defences.got () on the defence fails too, which makes me think the problem lies elsewhere. Will test shortly.

keneanung:
I didn't look into the defence.got() part yet, so I probably just didn't catch everything.

keneanung:
> directly calling defences.got () on the defence fails too

I just tested it and it doesn't fail for me (made the `defences` table public for testing and called it via `vlua`, `vshow defs` showed the defences afterwards just fine)

Issue #187, opened 2016-05-29
Enchanter issues


QwindorRousseau:
Current issue is that the enchanter uses room gmcp which was changed with the new areas for the cities to use for the enchanter. Updated script includes all the new rooms with the ouroburi to work properly

keneanung:
Changed script is at https://kobra.io/#/e/-KDjmHJR1x10CgTcikZ2, if anyone wants to make the PR before I get around to it...

PR #273, opened 2017-01-13
Update known rooms with ourobori

keneanung:
Fixes svof/svof#187


vadi2:
Looks fine to me, but I can't test if it's correct :)

keneanung:
Generally, the check for ourobori shouldn't be as apparent anymore, because it's done completely with GMCP now. But @QwindorRousseau gave me this a while back and had a bit of time on my hands, so why not?

Can you approve this PR?

Issue #186, opened 2016-05-27
"mount" command should exempt mounting figureheads.


jgh713:
Fairly self-explanatory. When someone attempts to mount a figurehead on their ship, it remembers their riding skill at "mount" and sets their mount to "figurehead". An exception should be added into the alias.

PR #274, opened 2017-01-13
Add exception of "mount figurehead" to riding tracking

keneanung:
When using "mount figurehead", svof would set your riding ability to "mount" and the mount to use to "figurehead". This includes an exception for that case now.

Also "board ship" would be triggered and sent unchanged, but the same things as above would have been done.

Fixes svof/svof#186


keneanung:
Tagging @jgh713 as this was your issue :wink:

PR #279, opened 2017-02-03
Revert "Fix namedb not resetting to correct colours after some echoes"

keneanung:
Reverts svof/svof#269


keneanung:
The actual bug was on the users side...

jgh713:
+1

This was legitimately annoying.

Issue #200, opened 2016-06-26
Inundated with phlegmatic humour detects slickness + loki instead of lethargy


ghost:
Inundating the phlegmatic humour shows the system getting slickness and loki, instead of (hidden) lethargy.

jgh713:
Gmcp tracks the proper afflictions (random from a pool of four, anywhere from one to four afflictions depending on severity of temper). This is still a bug, disabling the relevant alchemist trigger should fix it.

Kogannn:
To add to this I do not believe it is random afflictions given. If tempered from 1-3 it just gives lethargy when inundated. 4-5 give lethargy and anorexia. 6-7 give lethargy, anorexia, and slickness. And 8 tempers to phlegmatic give lethargy, anorexia, slickness, and weariness. Have tested.

jgh713:
Yeah that was outdated info. Is entirely deterministic nowadays.

PR #286, opened 2017-03-07
Removing hard-coded afflictions for phlegmatic inundate

jgh713:
Removing unknownany and hard-coded slickness from the inundate function, per #200

### DIFF
--- a/raw-svo.dict.lua
+++ b/raw-svo.dict.lua
@@ -992,17 +992,14 @@ dict = {
 
       -- gives various afflictions, amount of which depends on your humour level
       --[[
-        slickness always seems to happen
-        1-2: add 1 unknown
-        3-6: add 2 unknowns
-        7-9: add 3 unknowns
-        10: add 4 unknowns
+        1-2: 1 affliction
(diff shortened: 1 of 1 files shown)


raw-svo.dict.lua
```diff
@@ -1000,24 +1000,11 @@ dict = {
 
         anorexia 50% time
         slickness 8+
+		
+		Previous information is old. Now gives between one and four afflictions from the following: lethargy, slickness, anorexia, weariness.
+		Afflictions not hidden by gmcp, so removed from the inundated function.
```
keneanung: Why don't you remove the old information then?

jgh713:
I was told it was intended.

raw-svo.dict.lua
jgh713: Dunno.

keneanung:
Told by a player or was that a bug report response?

jgh713:
> Message #2786       Sent by Makarios
> 2017/3/11/23:29 Ah, that's intended, its not random. Inundate phlegmatic is entirely deterministic.

Issue #131, opened 2016-01-21
Serverside breathing not stopped when dragonforming


keneanung:
> Sent By: Kard On 1/21/12:21
> I was using "dragonform black" and svof didn't seem to pause breathing while the system was paused. I have serverside curing on. https://ada-young.appspot.com/pastebin/e9278e1c.

keneanung:
> Sent By: Keneanung On 1/21/16:57
>(To Kard): The log didn't show too much (for example no start of dragonforming) and I am no dragon myself so I can't confirm and this is the first report of the form. But I created an issue (https://github.com/svof/svof/issues/131) so I can look at it later. Jus to confirm: the (p) markers appear when you are dragonforming?

keneanung:
> Sent By: Kard On 1/21/17:17
> That's correct. On dragonform svo pauses until interrupted or complete. I had keepup breath on and it appears (maybe due to an issue with serverside curing and svo) that when I got balance back it tried to hold breath instead of noticing I'm paused.

keneanung:
> Sent By: Keneanung On 1/21/17:24
> (To Kard): And you told svo that you are using serverside (tn ss) right?

vadi2:
Yeah the log doesn't show the start of dragonforming, upon which Svof should've sent the command to serverside to pause serverside.

keneanung:
> Sent By: Kard On 1/21/20:33
> Correct. Serverside has been on for a while.

jgh713:
This seems to be fixed.

keneanung:
Will try to poke Kard about this.

jgh713:
Dragonforming only pauses svo local, not serverside. Needs to be fixed, should be relatively simple.

PR #288, opened 2017-03-07
Adding workaround for sending commands mid-alias.

jgh713:
Fix for #131, potentially others.


keneanung:
I'm not quite what the sequence is with the current implementation and how the changes fix that. Can you explain this?

jgh713:
@keneanung Dragonform is an alias. Part of the doaction code declares saves the old send functions, then declares the send functions to be empty so that nothing is sent when send() is called. Declaring force_send on load just allows access to the send function mid-alias.

This was a real bugger to track down, too.

Issue #165, opened 2016-03-22
Number of unknown afflictions is not correctly tracked


keneanung:
> Sent By: Jukilian On 3/21/13:02
> // When the 'unknown number of afflictions' display on the prompt (shown with ????) is cleared by touching tree and curing nothing, it doesn't seem that it resets the number of unknown affs you had back to 0. For example, I get hit with loki until I have ???. I touch tree, but I'd already cured the affs from their symptoms, so the system realises these unknown affs are gone, so it removes ??? from the prompt. However, next loki hit, ???? appears instead of ?. If that makes sense?

keneanung:
fixed with #289

PR #289, opened 2017-03-07
Clearing unknown affs on empty tree

jgh713:


jgh713:
Marking #165

vadi2:
Tree doesn't cure all affs though.

jgh713:
@vadi2

I was under the impression that pretty much all random affs were treeable. Is this not the case?

And if that isn't the case, why does treeable contain unknownany/unknownmental?

jgh713:
Also, should any2 and any3 treeable strategies consider multiple unknownmental or unknownany as treeable? At current, no matter how many of one you have, it's considered one.

jgh713:
Still have a few denizens to test, but as far as I can tell all unknownany are treeable, and all but one unknownmental affs are focusable.

keneanung:
@vadi2 @jgh713 what's the status on this?

vadi2:
Seems like a safe assumption to make though confusing the semantics even if they happen to align right now is not a good idea. I don't mind either way, just commented

Issue #221, opened 2016-08-09
Recognizing hypothermia/frozen combo


Eguzki27:
Svo erroneously assumes you're not frozen when you touch tree with hypothermia and frozen affs because tree doesn't cure anything. Ideally I want to not touch tree at all in this situation (unless I have other affs).

keneanung:
Hrm, from my search this may need a change in the tree scenarios directly. I'd also like to generalise this, so other afflictions that are blocked/buried below other affs are attempted to be cured. But I am not sure, if I can easily get this information in the scenarios.

PR #290, opened 2017-03-07
Adding checks for hypothermia when calculating/removing treeables

jgh713:
Per #221


svo (install the zip, not me).xml
```diff
@@ -54203,7 +54203,16 @@ end
 
 -- returns a list of affs that tree and other basic cures can get
 function svo.gettreeableaffs()
-  return table.n_intersection(svo.keystolist(svo.affl), svo.treecurables)
+  local a = table.n_intersection(svo.keystolist(svo.affl), svo.treecurables)
+  if svo.affl.hypochondria then
```
keneanung: Shouldn't this be `hypothermia`?

svo (install the zip, not me).xml
```diff
@@ -54203,7 +54203,16 @@ end
 
 -- returns a list of affs that tree and other basic cures can get
 function svo.gettreeableaffs()
-  return table.n_intersection(svo.keystolist(svo.affl), svo.treecurables)
+  local a = table.n_intersection(svo.keystolist(svo.affl), svo.treecurables)
```
keneanung: Why not using the `svo.treecurableswithhypothermia` here instead of removing the others manually later?

raw-svo.empty.lua
```diff
@@ -99,12 +99,14 @@ end
...
-
+empty.treecurableswithhypothermia = {"ablaze", "addiction", "aeon", "agoraphobia", "anorexia", "asthma", "blackout", "claustrophobia", "clumsiness", "confusion", "crippledleftarm", "crippledleftleg", "crippledrightarm", "crippledrightleg", "darkshade", "deadening", "dementia", "disloyalty", "disrupt", "dissonance", "dizziness", "epilepsy", "fear", "generosity", "haemophilia", "hallucinations", "healthleech",  "hellsight", "hypersomnia", "hypochondria", "illness", "impatience", "inlove", "itching", "justice", "lethargy", "loneliness", "madness", "masochism","pacifism", "paralysis", "paranoia", "peace", "recklessness", "relapsing", "selarnia", "sensitivity", "shyness", "slickness", "stupidity", "stuttering", "unknownany", "unknowncrippledarm", "unknowncrippledleg", "unknownmental", "vertigo", "voyria", "weakness", "hamstring", "skullfractures", "crackedribs", "wristfractures", "torntendons", "depression", "parasite", "retribution", "shadowmadness", "timeloop", "degenerate", "deteriorate"}
```
keneanung: I'd prefer to base off `treecurables` amd remove those uncurable with the technique you use in `gettreeableaffs`. The same with madness, actually...

svo (install the zip, not me).xml
jgh713: Bah, I pushed an earlier version of the changes.

raw-svo.empty.lua
jgh713: That's all well and good for hypothermia, maybe, but there's a helluva lot of things that madness makes non-treeable. Not quite as feasible.

svo (install the zip, not me).xml
jgh713: @keneanung svo.treecurableswithhypothermia wasn't public facing. Since we're changing the way that's handled though, I'm assuming you'd rather keep this the way it is.

keneanung:
This looks better because it's now easier to see what is blocked by which affliction. But I am not quite happy yet...

Maybe you can extract lines 108-115 of `raw-svo.valid.empty.lua` into an own function that is public and then use that in `svo.gettreeableaffs` instead of `svo.treecurables`?

raw-svo.empty.lua
```diff
@@ -97,17 +97,40 @@ empty.focus = function()
...
+      else
+        i = i + 1
+      end
+    end
+  end
+  return a
+end
+
+empty.tree = function ()
+  local a = gettreeableaffs(true)
```
keneanung: Why are you removing all afflictions here? Shouldn't the version without parameter be enough as it returns those that the system (thinks) you have, which are the only ones you need to remove?

jgh713: `if` not getall then
` -- stuff to remove afflictions we don't have

True returns all afflictions that are treeable. No input returns treeable afflictions that we have.

jgh713: @keneanung Oh, misread the question. Yeah, I guess it would be, that's just the way it was done before. Not sure what difference it would make, actually. Not sure if removeaff or the checks in that function are more intensive, but it's not at all noticeable and it's not a function that's called often enough for performance to be a real concern.  I can change it if you'd like.

PR #305, opened 2017-04-03
Changing hellsight checks from elm to valerian

jgh713:


keneanung:
Can you cure other valerian affs when hellsight and inquisition are present?

jgh713:
Seems so. So probably best to just add inquisition to the isadvisable for hellsight

PR #307, opened 2017-04-07
Adding -n flag to forcibly remove carriage returns

jgh713:
After some tired idiocy, a proper fix.

Using generate.lua on a windows machine creates a zip that's unuseable on other machines, and is just annoying in general. Optional flag to replace \r\n with \n.


jgh713:
Neeeeeeeeeeeeeevermind. Again. I fucking hate this thing.

PR #315, opened 2017-06-03
Added shrine identifier

vadi2:
It's unfinished because enemy/ally shrine are hardcoded. Should create some sort of mapping to namedb perhaps?

The script makes shrines visible in a room and also colors them by their health:
selection_084

Most shrine messages might be outdated, they're two years old.


keneanung:
I'm a bit hestitant to even review it if it is clearly stated that it's probably out of date and needs improvements (like not hardcoded enemy shrines). I don't think we should included with known deficits.

vadi2:
Sure.

PR #318, opened 2017-06-09
Shikudo Additions, Part Deux

natecox:
This PR is inspired by @jgh713 and credit should absolutely be given for getting me started. This PR changes a couple of specific items, detailed in commit 1f6aeaa.

Notably: `df` has been updated to remove the alternate monk path, and `vdefs` can now take shikudo forms in defup.

I'd appreciate a detailed code review here, even if this doesn't get merged in, so I can get an idea of any mistakes I've made.

### DIFF
--- a/raw-svo.defs.lua
+++ b/raw-svo.defs.lua
@@ -1799,42 +1799,42 @@ defs_data = pl.OrderedMap {}
     onenable = onenable_shikudo,
     def = "You are enacting the Tykonos form.",
     on = "You spin your staff in the opening sequence of the form of Tykonos, snapping into a ready stance.",
-    off = [[^You clumsily transition from the form of \w+ into the form of .+]]
+    off = [[^You clumsily transition from the form of \w+ into the form of]]
   })
   defs_data:set("willow", { 
     type = "shikudo",
     onenable = onenable_shikudo,
     def = "You are enacting the Willows shaken by the Wind form..",
     on = "Twirling your staff, you sink into the calm required for the form of Willows Shaken by the Wind.",
-    off = [[^You clumsily transition from the form of \w+ into the form of .+]]
+    off = [[^You clumsily transition from the form of \w+ into the form of]]
   })
   defs_data:set("rain", { 
     type = "shikudo",
     onenable = onenable_shikudo,
     def = "You are enacting the Willows in Rain Storm form.",
     on = "Dropping into a lower stance, you snap your weapon into an offensive position, tensing your muscles in preparation for the form of Willows in Rain Storm.",
-    off = [[^You clumsily transition from the form of \w+ into the form of .+]]
+    off = [[^You clumsily transition from the form of \w+ into the form of]]
   })
   defs_data:set("oak", { 
     type = "shikudo",
     onenable = onenable_shikudo,
     def = "You are enacting the the Live Oak form.",
     on = "Rising onto the balls of your feet, you prepare to begin the deadly form of the Live Oak.",
-    off = [[^You clumsily transition from the form of \w+ into the form of .+]]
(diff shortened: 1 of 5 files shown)


raw-svo.defs.lua
```diff
@@ -1770,6 +1770,74 @@ defs_data = pl.OrderedMap {}
...
+      "oak", 
+      "gaital", 
+      "maelstrom"
+    }
+
+    local fail_string =
+      "Removed %s from %s, it's incompatible with %s to have simultaneously up."
+
+    for _, morph in ipairs(shikudo_forms) do
+      if morph ~= newdef and svo["def"..whereto][mode][morph] then
```
vadi2: Might want to rename the morph variable

raw-svo.defs.lua
```diff
@@ -1770,6 +1770,74 @@ defs_data = pl.OrderedMap {}
...
+
+    return true
+  end
+
+  defs_data:set("tykonos", { 
+    type = "shikudo",
+    onenable = onenable_shikudo,
+    def = "You are enacting the Tykonos form.",
+    on = "You spin your staff in the opening sequence of the form of Tykonos, snapping into a ready stance.",
+    off = [[^You clumsily transition from the form of \w+ into the form of .+]]
```
vadi2: Not much point to the ending .+ if you're not capturing it

vadi2:
Looks fine from a design point of view! Nice work!

Just be careful about adding more locals, Lua has a limit of 255 of them and the reason things like codepaste as a table was to cut down on the amount of them.

raw-svo.defs.lua
```diff
@@ -1770,6 +1770,79 @@ defs_data = pl.OrderedMap {}
...
+  })
+  defs_data:set("willow", { 
+    type = "shikudo",
+    onenable = onenable_shikudo,
+    def = "You are enacting the Willows shaken by the Wind form..",
```
keneanung: Are there two periods in the original def line?

raw-svo.dict.lua
```diff
@@ -89,6 +89,80 @@ local dict_smoke_def = {}
...
+  }
+
+  for _, stance in ipairs(stances) do
+    defences.lost(stance)
+  end
+
+  defences.got(new_stance)
+end
+
+local tekura_stance_isadvisable = function (new_stance)
```
keneanung: This doesn't seem to be used for tekura forms only. Can you update the function name accordingly? Same with `shikudo_form_advisable` below.

keneanung: Ignore the shikudo part of the comment...

raw-svo.setup.lua
```diff
@@ -309,6 +309,38 @@ signals.gmcpcharvitals:connect(function()
...
+  if gmcp.Char.Vitals.charstats then
+    for index, val in ipairs(gmcp.Char.Vitals.charstats) do
+      local stance = val:match("^Stance: (%w+)$")
+      local form = val:match("^Form: (%w+)$")
+
+      if stance then
+        me.path = "tekura"
+        me.form = nil
+        me.stance = stance:lower()
+        sk.ignored_defences['shikudo'].status = true
```
keneanung: Do you need to unignore tekura here?

raw-svo.defs.lua
natecox: Nope, this is updated in the current push.

raw-svo.dict.lua
natecox: Functions have been renamed to be more appropriate.

raw-svo.setup.lua
natecox: Yeah, that's probably a good idea. I've updated both sections to swap appropriately.

PR #325, opened 2017-10-13
Changed trigger so other people inking wont cause you to touch your tattoos.

nylian:
Changed trigger so other people inking wont cause you to touch your tattoos.

### DIFF
--- a/svo (inker).xml
+++ b/svo (inker).xml
@@ -83,7 +83,7 @@
                 <name>Finished inking the tattoo</name>
                 <script>svo.ti_finishedinking()</script>
                 <triggerType>0</triggerType>
-                <conditonLineDelta>1</conditonLineDelta>
+                <conditonLineDelta>0</conditonLineDelta>
                 <mStayOpen>0</mStayOpen>
                 <mCommand></mCommand>
                 <packageName></packageName>
--- a/svo (inker).xml
+++ b/svo (inker).xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE MudletPackage>
-<MudletPackage version="1.0">
+<MudletPackage version="1.001">
     <TriggerPackage>
         <TriggerGroup isActive="yes" isFolder="yes" isTempTrigger="no" isMultiline="no" isPerlSlashGOption="no" isColorizerTrigger="no" isFilterTrigger="no" isSoundTrigger="no" isColorTrigger="no" isColorTriggerFg="no" isColorTriggerBg="no">
             <name>svo Tattoo Inker</name>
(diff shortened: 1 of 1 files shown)


svo (inker).xml
```diff
@@ -93,11 +93,9 @@
                 <colorTriggerFgColor>#000000</colorTriggerFgColor>
                 <colorTriggerBgColor>#000000</colorTriggerBgColor>
                 <regexCodeList>
-                    <string>As the</string>
```
keneanung: You should leave this trigger as a `multiline` trigger. It is an optimisation because RegEx triggers are relatively expensive performance wise. Begin of line triggers are the cheapest kind of trigger and as such act as a filter in this case.

nylian:
Thanks Keneanung.  Trigger costs in mudlet are pretty new to me, would we want to do multiline with 'As the' as a begin line AND have the regex line?

Is there some where I can review costs like this? Makes me want to rebuild my system.

keneanung:
Yes, we want to have a multiline trigger with both a begin of line trigger as 1st condition and regex as the second condition. That way the regex is shielded: https://forums.mudlet.org/whatever-f12/doing-shielded-regex-differently-t1441.html

This is most useful if the shielded trigger is expected to match relatively rarely. If the regex matches every 2nd line, the shielding simply adds to the actual trigger.

The order goes somewhat like this (though no formal profiling was done. It's deduced from the kind of operation that has to be done):
1. Begin of line
2. Exact Match
3. Substring
4. Perl regex
5. Lua function

I am not sure where color triggers fit in there, but I think it's somewhere in the order "substring".

nylian:
Requested changes have been made. Thanks for all the help with git, Keneanung!

svo (inker).xml
```diff
@@ -83,7 +83,7 @@
                 <name>Finished inking the tattoo</name>
                 <script>svo.ti_finishedinking()</script>
                 <triggerType>0</triggerType>
-                <conditonLineDelta>0</conditonLineDelta>
+                <conditonLineDelta>1</conditonLineDelta>
```
keneanung: The Delta should be `0` (they should be the same line). As it is now if a line starts with "As the " but is not the ending line for tattooing, it will try to match this line to the regex (can't be avoided) and the next one.

nylian:
Fixed and pushed!

Kogannn:
Since our regex is now...

`^As the \w+ tattoo's shape is codified on the .+ of yourself, the last of the ink leaves the air and you pause for a moment to admire your handiwork\.$`

We run into a new issue where we have to manually unpause after inking someone else since yourself is hardcoded in. However if we regex the yourself then we go back to touching tattoos whenever we ink someone else.  Will think on a solution for this. Can likely do something with matching it where if its not "yourself" then we just don't touch the tattoo.

PR #327, opened 2017-11-12
Added @battlerage prompt tag to documentation.

Kogannn:
@battlerage prompt tag added to 'Setting a custom prompt' section. To be found right below @power now.


keneanung:
Thank you very much for the update and congratulations for your first contribution!

I am not sure when this change will be life as documentation updates get pushed with new releases, but I don't want to nag people with a new release (downloading and installing the update) for a doc update only...

PR #334, opened 2018-01-23
add_post_emit changes

CorTheWin:


svo (setup, misc, empty, funnies, dor).xml
```diff
@@ -865,7 +865,7 @@ if not svo.systemloaded then
     if not conf.curemethod or conf.curemethod == 'auto' then
       conf.curemethod = 'conconly'
     end
-  end)
+  end, 'post emit systemstart')
```
vadi2: This should be 'adjust cure method' - otherwise people will get the wrong idea when copy/pasting and call something else 'post emit systemstart'

vadi2:
Thanks! Will put out an update with this soon.

PR #349, opened 2018-02-06
Post_emit save the name

CorTheWin:


vadi2:
Yep, I think that was the bit you were missing!

PR #362, opened 2018-08-31
Remove website scraping ndb.getinfo()

ndsamuelson:
the file download is no longer allowed.


ndsamuelson:
Didn't realize api.achaea.com/characters was even a thing until I submitted this. closing, as that appears to be a much better option

PR #363, opened 2018-10-01
Add docker-related config for build

aaronjoslyn:
Hello,

I had a few issues getting luarocks to play nice on OSX, so decided to throw this together.

If you have docker and docker-compose installed, you can use e.g. `CLASS=magi docker-compose up` to generate for Magi. I haven't been using svof for long, so hopefully I haven't missed anything :)

Thanks for svof!


aaronjoslyn:
I'd have liked to add more detail but install instructions for Docker can vary quite a bit depending on platform :S

PR #365, opened 2018-12-22
Remove upload to pathurs.com

keneanung:
This is an old remnant that doesn't seem to work anymore.


pathurs:
Ah! I am glad you don't rely on my website anymore. I am about to reset it and was checking if this was still required.

PR #329, opened 2018-01-06
in-client & multiclass Svof [WIP]

vadi2:
This will enable the system to be fully inside Mudlet and completely editable. I'm hoping the community will pick up on updating the system with this (massive) change in.

Instructions on installing: http://forums.achaea.com/discussion/comment/390947/#Comment_390947

Multiclass is included.

Todo list:
- [x] remove debug preprocessor calls
- [x] remove class and skill preprocessor calls
- [x] remove all other code generation processor calls
- [x] make necessary locals public
- [x] ensure `svo.me.class` checks against class with uppercase
- [x] convert as many doublequotes to singlequotes as patience allows
- [x] migrate all of the functions to use the Svo namespace
- [x] enable skill detection for the first run using gmcp
- [x] check for final `luacheck` errors
- [x] migrate core files into Mudlet
- [x] change LuaNotify to allow signal overrides on connect()
- [x] migrate addons into Mudlet
- [x] move default prios to Muldet
- [x] fix ndb help to work
- [x] update setup instructions
- [x] delete outdated and irrelevant files
- [ ] Achaean community: test the changes
- [ ] Achaean community: provide list of new default curing and defup priorities

Test list. Alchemy can be swapped in for where Concoctions is mentioned:
- [x] blind/deaf are correctly recognised as afflictions when they're not on keepup/defup for Apostates
- [x] blind/deaf are correctly recognised as afflictions when they're not on keepup/defup for non-Apostates
- [x] triggers for getting afflicted with "skullfractures", "crackedribs", "wristfractures", "torntendons" work
- [x] bleeding diag trigger works
- [x] at least one other non-bleeding diag trigger works
- [x] list of serverignore/cadmusaffs/curelist saves and loads fine
- [x] tree and skill (like bloodboil) cure triggers for "skullfractures", "crackedribs", "wristfractures", "torntendons" work
- [x] triggers for application of "skullfractures", "crackedribs", "wristfractures", "torntendons" work
- [ ] restoration curing something other than the intended affliction works
- [ ] regeneration curing something other than the intended affliction works
- [x] focus curing something other than the intended affliction works
- [x] smoke curing something other than the intended affliction works
- [x] herb curing something other than the intended affliction works
- [ ] humour getting cured other than the intended affliction works
- [ ] parry works for blademasters with their sword sheathed
- [ ] guarding works for monks without anything wielded
- [x] parry doesn't go off for everyone else while no weapon is wielded
- [x] writhing works for non-bards
[...]

PR #378, opened 2019-07-02
Added water/fire/air/earth/psion affs to main

jmclicquennoi:
Should have every new affliction added to the compiled version. Mostly this is just to have it here, while we work on either updating this further or moving to in client.

It does not have all of the trigger lines, but the affs themselves are there, and the dict structure is nearly the same as the current in client's.


vadi2:
What happened to `svo (install the zip, not me).xml` - every single line is changed. Did you change the line endings?

keneanung:
Looks like it. If you switch to "ignore whitespace changes", the number of changed lines goes down to 2k (which is still a lot)

QwindorRousseau:
Has been added or will be added to #329 which will be the next official release

PR #383, opened 2019-07-06
Auto Reject fixes

### LINKED ISSUE #371: Autoreject issues (double lust & missing trigger).
The autoreject logic will not be invoked when looking at ALLIES if the luster also has you lusted. This is suddenly more likely due to a new card that was added (Travian).

The trigger "Lust" under "Tarot" must have its second regex updated to account for "(M)" at the end:

> You feel an unusually strong lust for Seelie (M).

Suggestion is the second regex be updated to 

`^You feel an unusually strong lust for (\w+)(\.| \(M\)\.)$`

Additionally, the trigger "svo cure lovers nobody" must be updated to account to the new format when rejecting someone who is not lusting you:

> Why would you reject Seelie?

Suggestion is to add the following regex to the trigger:

`^Why would you reject (\w+)\?$`

Thank you.

DazedxJoker:
This should fix #371, and make it so Svof will reject lust if it is mutual.


vadi2:
Make sure to start with the in client branch, not the master one 😁

PR #390, opened 2019-07-06
Defense list fix

QwindorRousseau:
Simple fix that was just overlooked.


QwindorRousseau:
Nevermind. Not understanding problem correctly

Issue #295, opened 2017-03-14
Remove inking alias from tattoo inker


keneanung:
> Message #831        Sent by Lynara
2017/3/08/22:17 The 'inking' alias and outrifting such should be removed from svof, Achaea does that 
natively now and it gets all broken when you try to use the svof version. triggers are still good 
for it, because they pause the system and all that.

nylian:
How do we want to adjust this? Personally I'd rather take everything SVO related out, except for the pause -
 including the "be still" comment and the retries.  Getting caught in an attack while trying to tattoo is frustrating and potentially robs a bunch of inks.

keneanung:
Another option would be to allow the configuration of "max retries" and allow the inking process to get interrupted manually. I'm a bit careful to remove a feature as some people might consider it useful.

QwindorRousseau:
Fixed with #394

PR #394, opened 2019-07-07
Inker update

QwindorRousseau:
-The game natively pulls the inks from the rift when you start a tattoo so I edited the outr of inks to keep the spam down and the potential to lose inks if something happens while you are inking.


QwindorRousseau:
Will fix #295 for outring inks as part of the system. Keeping the module though as it can batch tattoos together for when you need it.

PR #398, opened 2019-07-08
Priest Changes

K1ngSpad3z:
Need someone to check prayer balance, as of now the prayers get spammed for the blessings defences


K1ngSpad3z:
With this pull:

Healing has been removed
Zeal has been added
Prayer balance has been added

To finish:

benediction function to use benediction over individual blessings
add zeal affs into system

To test:

Prayer balance

PR #418, opened 2019-07-14
Limb Counter Addon Changes

K1ngSpad3z:
Moved all the limb counters from each class into one addon for accessibility.

Added in a trigger to reset limbs based off salve balance

Will need someone to help with the coding for:

Meta
Magi
Monk

as I do not have those classes to test or manage the triggers for it. Everything else works though currently


svo (limbcounter).xml
```diff
@@ -0,0 +1,1920 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE MudletPackage>
+<MudletPackage version="1.001">
+	<TriggerPackage />
```
QwindorRousseau: should be `<TriggerPackage>`

Issue #205, opened 2016-07-08
Relighting and svo.sendc


jgh713:
Svo doesn't properly recognize emptying pipes as a command that requires balance, so when used in conjucnction with sendc, relighting doesn't work properly, and just outrs a lot of whatever plant/mineral and spams attempting to empty/relight.

Not sure where or how one would address this, so if someone could take a look, that'd be great.

keneanung:
It may be enough to replace the send in https://github.com/svof/svof/blob/master/raw-svo.rift.lua#L664 and following by a `doaddfree()`. (https://github.com/svof/svof/blob/master/raw-svo.rift.lua#L707 as well)

QwindorRousseau:
@svof/developers Again lets take a look and see if this is still an issue and if it is then above information should make this an easy fix.

PR #421, opened 2019-07-16
Pipe Balance fixes

QwindorRousseau:
as per #205 this fixes the issue with emptying pipes not respecting balance. Did the included changes suggested and it appears to be working correctly now.


vadi2:
I wouldn't use the do queue for it - remember it's meant for the players use, not the systems use. That's why you never see the system using it itself. If the system is paused for example and the player clears the do queue, they wouldn't expect the system to break and it would break here...

PR #428, opened 2019-07-19
Serverside Deafaff and Blindaff fixes

K1ngSpad3z:
serverside now correctly removes and defs blindness and deafness


K1ngSpad3z:
We need to update the svo prio list to move deafaff in line with blindaff so they can both be set to 25 on the in game aff priority list

PR #440, opened 2019-07-22
Pipe Changes

K1ngSpad3z:
Currently serverside curing is set up to automatically handle pipes. This commit has the following added:

Fixed the dict to tell svo not to worry about pipes (relighting and refilling) if serverside is turned on

Fixed the balance changes on refilling and relighting pipes

Fixed taking out an herb after trying to put it in the pipe first (was refill/outr and its now outr/refill)


K1ngSpad3z:
Fixes Issue #205

PR #441, opened 2019-07-22
Serverside error for syncing slowcuring

### LINKED ISSUE #436: Error during Serverside Priority Sync
In order to reliable create this error: Turnoff Serverside curing and force a classchange.

image

I believe the error is caused when sk.notifypriodiffs() runs into these when comparing differences.

image

DazedxJoker:
Fixes #436


vadi2:
why does afterstate and beforestate even have this balance? seems like a waste to compute it and then ignore last minute, no

DazedxJoker:
> Why do afterstate and beforestate even have this balance? seems like a waste to compute it and then ignore last minute, no

It's in the list of slowcuring priorities for some reason, I'm not sure.

DazedxJoker:
Found a better solution, I believe.

Issue #431, opened 2019-07-20
Update unnamable


antar5:
The line for unnamable needs updating. Can remove the unnamable visions/speak start line, and add this line instead (when unnamable completes).

^You realise with horror that the eldritch utterence has caused (\d+) insanities to descend upon you\.$

This line will specify in the first match whether you get 1, 2, or 3 affs randomly selected from a pool of stupidity, confusion, and dementia. The delivery is smart, so if you have any of these afflictions beforehand unnamable will afflict with any remaining affs in the pool when it completes. This also means that if you are hit with a 3 unnamable, you will always know what the afflictions are.

QwindorRousseau:
fixed with #458

PR #458, opened 2019-07-27
Unnameable logic

QwindorRousseau:
Added in a bit more logic with the occultism unnameable skill line. Could probably use a bit more work for logic but this will do the intended purpose for the time being. Also added the enchanter module into the loader as I had forgotten it somehow.


QwindorRousseau:
for #431

PR #459, opened 2019-07-27
fix(dragonbreath-func): cleanup code for dragonbreath, use function from triggers

sheepdog-mcfry:
Add unpause serverside function for switching class for class changes that require pause (dragon, elemental)


svo (aliases, triggers).xml
```diff
@@ -21192,21 +21192,9 @@ svo.parseplistend()</script>
...
-   ["red"] = 'dragonfire',
-   ["black"] = 'acid',
-   ["silver"] = 'lightning',
-   ["gold"] = 'psi',
-   ["blue"] = 'ice',
-   ["green"] = 'venom'
-}
-
-if not t[multimatches[2][2]] then return end
+					<script>svo.valid.dragonformingcolour(multimatches[2][2])
```
QwindorRousseau: There is no reason to add this in there. The colour for breath weapon gets set on the middle dragonforming trigger hence the reason for the start dragonforming so it will recognize the second trigger and then complete with the third one.

Issue #427, opened 2019-07-18
Dragon breath tracking


QwindorRousseau:
The lines for summoning the breath weapons were changed to show the specific types of breath along with the lines when it is used. It just needs to be added so it will track it correctly.

QwindorRousseau:
fixed with #460

PR #460, opened 2019-07-27
Dragonbreath tracking fix

QwindorRousseau:
The functions on the triggers were trying to access a definition in the dictionary for the defences that was never defined so the check for the completion would never finish and update.


QwindorRousseau:
for #427

PR #463, opened 2019-07-27
quick stun fix

QwindorRousseau:
Looks like it was trying to set an old version of the function which isn't really used anymore and with this I am no longer getting the LUA error to my console


QwindorRousseau:
for #463

PR #464, opened 2019-07-27
Fixes for Flail Staff

K1ngSpad3z:
The action wasn't being recognized so svo thought it was an illusion. This should fix it


vadi2:
Didn't test, but this makes sense.

K1ngSpad3z:
Was the same issue for dragon breath not capturing correctly. Just gonna put it through and if it errors out again we can check it after beta release

vadi2:
Actually you're missing a comma there, oops

PR #474, opened 2019-07-29
Temptrigger tables

sheepdog-mcfry:
I also removed my temp_defs until the issue is resolved to not create multiple points of failure in terms of defenses.

This doesn't fix the issue that we are having in regards to defs stalling on system class change, but it prevents a build up of infinite temp triggers from class changing.


svo (alias and defence functions).xml
```diff
@@ -4140,13 +4140,15 @@ signals.systemstart:connect(function ()
 --Empty the def_type table on startup and class change to prevent duplicate defences.
  defences.def_types = {}
  
+ --organise tempTriggers in a table to delete on class change
+ svo.tempDefTriggers = {}
```
vadi2: System uses lowercase with spaces optionally, not camelcase

svo (alias and defence functions).xml
```diff
@@ -4156,71 +4158,71 @@ signals.systemstart:connect(function ()
 
     if defdata.onr and type(defdata.onr) == 'table' then
       for _, pattern in ipairs(defdata.onr) do
-        tempRegexTrigger(pattern, 'svo.defs.got_' .. sk.sanitize(defname) .. '()')
+        table.insert(svo.tempDefTriggers, tempRegexTrigger(pattern, 'svo.defs.got_' .. sk.sanitize(defname) .. '()'))
```
vadi2: mytable[#mytable+1] = is faster, and we're creating a lotta triggers

svo (install me in module manager).xml
```diff
@@ -7174,36 +7174,29 @@ end</script>
...
+
 		
   if svo.me.class ~= newclass then
     local oldclass = svo.me.class
 
+    -- kill temporary defense triggers for old class
+    for k,v in pairs(svo.tempDefTriggers) do
+      killTrigger(v)
+    end
+
```
vadi2: Clear the able after killing? Else it'll keep growing

vadi2: That won't clear it as the key is K, how about resetting the table after the loop?

PR #482, opened 2019-08-02
qwhom fix

QwindorRousseau:
The lua error for wrapLine() was due to it being a two argument function when the trigger was only calling one argument which was in the wrong place due the function also requiring the console it was trying to send it to.


vadi2:
To be honest this could be a potential bug in Mudlet because `main` argument is always supposed to be optional :O I'd report it there.

QwindorRousseau:
The mudlet wiki doesn't state that for the function. I just checked to confirm it.

Issue #527, opened 2019-10-01
NDB won't trigger honours'ing the next name when you can sense everyone.


ndsamuelson:
In the example output below of qw/qw2, apparently I can sense everyone at the moment, and the end of the trigger doesn't fire because it's looking for "Plus another # that you cannot fully sense" or whatever it says. (Sorry, I've been asleep a long time and just happened across this.)

>Aalya, Aeka, Ainly, Allene, Amanda, Ammar, Anwen, Apsethus, Archaeon, Ariza, Armali, Astarod,
Ayaris, Azarae, Baldazar, Boosteya, Colgano, Curtis, Darrod, Dhaia, Edotlina, Elyra, Esti, Ethoas, 
Farrah, Faur, Ferrn, Fhenir, Frances, Gavriil, Gilliam, Havyn, Hirst, Iboyr, Ilsabet, Imia, Ipomoea, 
Issam, Iulian, Jadys, Jamia, Jarrod, Jaxster, Jinos, Kaburia, Kassian, Kazuya, Keorin, Keyo, Kitiara, 
Kog, Kriex, Kuririn, Laedha, Layta, Lenn, Lorath, Lorei, Marcella, Martien, Mathonwy, Mezghar, 
Micaelis, Mina, Mizik, Morthif, Mycen, Myna, Naelle, Namielle, Nhakra, Pandora, Penwize, Profit, 
Rackham, Rane, Ranok, Rhavyn, Rom, Ryzan, Sabetha, Sanochpe, Santar, Shandrala, Shirszae, Sothantos, Syrei, Tahquil, Tarvius, Taryius, Tashar, Teine, Tesha, Thaisen, Thiev, Trendric, Twilight, Tysel, Uilien, Uvian, Vastar, Xallax, Xchuul, Yina, Zadkiel, Zekiel, and Zenui.
(107 total).

ndsamuelson:
seems that pull request only fixed it temporarily. I'll investigate it a bit more.

ndsamuelson:
interesting note though, if you have a queue of names that are not honoursing, if you keep doing cwho they'll all cycle through heh.

DazedxJoker:
Pretty sure this issue was already fixed in the in-client build

ndsamuelson:
I'm using the in-client build currently. fresh installed yesterday. But i'll fiddle. Thanks!

ndsamuelson:
seems my build was behind. sorry everybody. heh.

PR #528, opened 2019-10-02
NDB qw/qw2 regex fix.

ndsamuelson:
fix (i believe anyway) for issue #527


ndsamuelson:
This was only a temporary fix it seems. More investigating will be needed.

PR #565, opened 2020-01-05
Update svo (namedb).xml

### LINKED ISSUE #560: Remove leading `> ` from qw alias output
The leading `> ` prior to the qw output is unnecessary and kind of annoying. Would be great to remove it!

PhoenixCodes:
Closes #560

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -942,8 +942,7 @@ elseif ndb.qwtype == "show infamous" then
 end
 
 commit()
-disableTrigger("NameDB qw")
-echo(" Fired")</script>
+disableTrigger("NameDB qw")</script>
 					<triggerType>0</triggerType>
 					<conditonLineDelta>1</conditonLineDelta>
 					<mStayOpen>0</mStayOpen>
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -752,15 +752,12 @@ temp_name_list = {}
 
 local selectString, getFgColor, format = selectString, getFgColor, string.format
 
--- selectString not working for first name by default - fix with this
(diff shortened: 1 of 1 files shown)


svo (namedb).xml
```diff
@@ -945,7 +942,8 @@ elseif ndb.qwtype == "show infamous" then
 end
 
 commit()
-disableTrigger("NameDB qw")</script>
+disableTrigger("NameDB qw")
+echo(" Fired")</script>
```
QwindorRousseau: Need to take your echo check out

svo (namedb).xml
```diff
@@ -959,7 +957,7 @@ disableTrigger("NameDB qw")</script>
 					<regexCodeList>
 						<string>.*</string>
 						<string>1</string>
-						<string>^(?:Plus another (\d+) whose presence you cannot fully sense(?: \(\d+ total\))?\.|\((\d+) total\)\.)$</string>
```
QwindorRousseau: I did this change with the original issue where it was not working when there were no hidden people so this should not need to be changed.

PhoenixCodes: The changed line works both ways (tested it to verify that.) and is slightly more optimized from a regex standpoint. I removed unnecessary captures, and it doesn't have to run the entire search twice since a part is optional.

svo (namedb).xml
PhoenixCodes: Done

svo (namedb).xml
QwindorRousseau: My bad. Forgot there was a ninja change they did on me after I had kept the capture in the first place.

Issue #563, opened 2020-01-05
Class switching (and lesserform/Dragonform) can cause issues


PhoenixCodes:
System seems to get confused with which defences are up, which need to go up again, which fell due to class change, etc. For the in-client, we need to have the system not get screwed up due to a class change or dForm.

PR #568, opened 2020-01-07
Retain general defences through classchange

DazedxJoker:
System now recognized that general defences weren't lost on classchange


DazedxJoker:
Fixes #563

PR #575, opened 2020-01-08
Code to make updater update, plus version increment to 34 from 33.4

PhoenixCodes:
This will allow auto-updating to work... It's clunky and somewhat dirty, but it does do the job for now.

### DIFF
--- a/svo (install me in module manager).xml
+++ b/svo (install me in module manager).xml
@@ -6380,42 +6380,29 @@ stupidityCheckTable = {
       if svo.announceupdates == "checking" then svo.echof("A new Svof is available and has been downloaded for you! You're on %s, latest is %s.", svo.version, s:trim()) end
       svo.showupdatereminder(writtenversion)
     end
-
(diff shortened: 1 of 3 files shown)


svo (install me in module manager).xml
```diff
@@ -6425,11 +6412,12 @@ end</script>
 				<name>Check for updates</name>
 				<packageName></packageName>
 				<script>local downloadfolder = getMudletHomeDir().."/svo/downloads/"
+local baseUrl = "https://github.com/svof/svof/raw/in-client-svof/"
```
TheLastDarkthorne: Shouldn't this point out to the release zipped version that Travis CI builds instead of the raw files in the github repo?

PhoenixCodes: No. Quick and dirty and is using individual xmls. In the future I'd hope we can do individual file versioning, instead of updating the whole system with every update. This is a step towards that.

K1ngSpad3z:
Tested, and now squashing!

PR #591, opened 2020-02-02
Adding file versions

PhoenixCodes:
5 modules did not have script packages, and so had no file versioning info. Added script packages and file versioning to them.

Added workarounds for setModulePriority

Added help URLs to the modules.

Adjusted svo (install me in module manager) to handle scriptable installing/uninstalling, and to do one-click uninstall.

### DIFF
--- a/svo (install me in module manager).xml
+++ b/svo (install me in module manager).xml
@@ -7417,6 +7417,6 @@ end</script>
 	</ScriptPackage>
 	<KeyPackage />
 	<HelpPackage>
-		<helpURL></helpURL>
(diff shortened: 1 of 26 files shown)


svo (install me in module manager).xml
```diff
@@ -7272,13 +7271,13 @@ end</script>
 	-- sync this module too
 	if syncModule then syncModule('svo (install me in module manager)') end
 end
-svo.install_other_modules()
```
PhoenixCodes: This line would never do anything, as the function will only complete when passed a modulename as the second argument.

PR #599, opened 2020-02-18
Update svo (install, config, pipes, rift, parry, prios).xml

PhoenixCodes:
Remove save error when nobody has logged in


ndsamuelson:
Why do you always break Travis @PhoenixCodes  😃

Issue #603, opened 2020-03-17
Inline ASCII bars crash


wysvaine:
The trigger for bar health on assess is susceptible to illusions - an illusion of 99999/1 health or similar will cause mudlet to crash as it tries to generate an impossibly large bar.

![image](https://user-images.githubusercontent.com/62292551/76858999-d89a5a00-682e-11ea-837a-bafe0baa6b0b.png)

[inline bar health xml.txt](https://github.com/svof/svof/files/4343775/inline.bar.health.xml.txt)

ndsamuelson:
Dang why did I never try that!? Good show. 👍 @wysvaine

vadi2:
The trigger is off by default in Svof, thankfully :smile:

PR #604, opened 2020-03-17
Let's just cap that

DazedxJoker:
Capping the visual health bars at 180%, (I picked that because that's how big I could go to keep it on 1 line.)


DazedxJoker:
Fixes #603

PR #606, opened 2020-04-02
Pre-login prevention

DazedxJoker:
Prevents the system from doing more things before logging in.
Also, adjusts when the system thinks it is logged in.

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -2265,7 +2265,7 @@ end
 svo.sendc = sk.sendqueuecmd
 
 function svo.sk.dosendqueue()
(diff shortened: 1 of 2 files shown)


vadi2:
Hey I have a cool idea -

https://github.com/svof/svof/blob/c15b782b52518b14d77d54bbc001c9f3db003fb0/raw-svo.skeleton.lua#L833 is where the system dynamically changes what kind of checks should be done - ones that send everything asyncronously or syncronous ones. For aeon curing, obviously.

What if you add a logging_in check to just call an empty function?

Would that break things? It would certainly remove all these checks we had to add.

DazedxJoker:
It looks like that achieved the same effect.

PR #616, opened 2020-04-21
clickable url fix

druuimai:
Ensure that qt don't use file://; instead, it will use http:// by default if ftp://, http(s):// or telnet:// haven't been used.


PhoenixCodes (approved):
This works, probably adds a layer of security to URLs and fixes some malformed ones.

Issue #617, opened 2020-04-21
Autorewield spam


DazedxJoker:
**Issue:**
![image](https://user-images.githubusercontent.com/14912622/79919980-57218300-83fd-11ea-8742-81aaa9e7a312.png)

If you send a wield command while you're already wielding something and it causes you to unwield the first item. If **vconfig autorewield** is on, it will cause a loop of rewield spam.

**Steps to reproduce:**

1. vconfig autorewield on
2. wield a 2h weapon (ie. fishing pole)
3. wield anything else.

**Expected result:**

Svof should recognize that wield command was send by either the user or itself and not trigger an rewield.

PR #618, opened 2020-04-26
Fix for issue #617: Autorewield spam

shasta-schmuck:
Added a line to the rewield check with the actual triggering text, and it seems to have corrected the spam.


PhoenixCodes:
Could you please fix which folders/triggers are enabled and not so they match the older version? Thanks!

shasta-schmuck:
Sorry about that! Corrected.

PR #620, opened 2020-05-03
Inclient mention

DazedxJoker:
Just adding a mention and a link to the inclient in the readme.


DazedxJoker:
@vadi2  You might have to merge that, I don't think the checks will finish.

Issue #625, opened 2020-06-29
Sylvan Viridian Limbcounter


VinceresEbonheart:
The Viridian Limbcounter is placed into the Meta limbcounter folder, but Meta is for Sentinels and Druids. Viridian is Sylvan and it should be moved into it's own respective folder.

Also, the coding is incorrect. 

Trigger:

You command the razor-edged thorny vines around you to lash out and rend the
^You command the razor-edged thorny vines around you to lash out and rend the (.+) of (\w+)\.$

Response:

local limbs = {"head", "torso", "right arm", "left arm", "right leg", "left leg"}
if table.contains(limbs, multimatches[2][2]) then
  local hit = svo.lc_break_at / 4
	hit = string.format("%2.1f", hit)
	hit = tonumber(hit)
	svo.lc_hit(multimatches[2][2], multimatches[2][3], hit) --------------- svo.lc_hit(where, who, howmuch)
end

====

it should be ....

svo.lc_hit(multimatches[2][3], multimatches[2][2], hit) --------------- svo.lc_hit (who, where, howmuch)

Issue #622, opened 2020-06-08
Earthlord Limb Counter


VinceresEbonheart:
I may be incorrect with my logic, but it seems the earth lord limb counter isn't calculating shaping properly. 

When someone powderises a limb, the amount of limb damage is based on how many levels of shaping the person has prior to the attack. For example, if I have level 1 shaping, it should do 1 limb damage, then elevate my shaping to level 2. 

However, it looks like the limb counter is elevating the level of shaping PRIOR to assessing how much limb damage the powderise is doing. For example, if I have level 3 shaping and I powderise a limb, it _should_ elevate the limb count by 1 (since I only had level 3 shaping when I attacked). Instead, what it's doing is raising my level of shaping to 4 which would then mean that it's saying my powderise is elevating by 2 hits, which is not correct.

This is what it looks like in the code:

local shaping = string.match(gmcp.Char.Vitals.charstats[3], "%d+")
shaping = tonumber(shaping)
if shaping == 4 then
	svo.lc_hit(matches[3], matches[2], 2)
elseif shaping <= 3 then
	svo.lc_hit(matches[3], matches[2], 1)
end

Instead I think it should be rearranged to look like this:

if shaping == nil or shaping <= 3 then
  svo.lc_hit(matches[3], matches[2], 1)
elseif shaping == 4 then
  svo.lc_hit(matches[3], matches[2], 2)
end
shaping = string.match(gmcp.Char.Vitals.charstats[3], "%d+")
shaping = tonumber(shaping)

Also, it looks like SVO says it should break with 6 hits, but earth lord breaks at 5 (at least when I do it without artefacts).

VinceresEbonheart:
Only problem I can foresee with the way I have it is that shaping is also increased by Pulverise as well. So a separate trigger might need to be made to ensure it tracks shaping that way too.  "You rain a flurry of blows down upon \w+ with your stone fists."

VinceresEbonheart:
No stale.

K1ngSpad3z:
Won't need the trigger for Pulverise. Going to have to add in a variable to track shaping on gmcp output and use the above trigger like that. the way the limbcounter works is that it does the calculations on next prompt, so it won't need to update the shaping count after the hit is sent through. I will take a look at this when applying a few other changes @VinceresEbonheart

VinceresEbonheart:
[...]

Issue #610, opened 2020-04-12
Monk Limbcount


VinceresEbonheart:
I had updated to the new in-client, and my Monk Limbcount doesn't appear to be working anymore. In previous versions, it used svo.ml_hit to track monk limb damage and used the @ml_prompttag for display. I know there was a previous ticket that said the prompttags were updated and to use the new updated ones.

I'm currently using the @lc_prompttag which does show on my display, but svo doesn't seem to be tracking any monk limb hits. Looking through the code, my monk limb tracker triggers still seem to be using svo.ml_hit -- when I search for the corresponding function, the function doesn't exist. I'm assuming it needs the updated to use the lc_hit function, but when I tried to change the triggers to use it, it didn't work correctly.

vadi2:
I think @K1ngSpad3z merged them all into one limbcounter - maybe there's a different tag now. Do the docs say anything about this?

VinceresEbonheart:
I thought that was the case as well. Using @lc_prompttag works fine with my magi and druid, but it doesn't look like monk (tekura) was updated? Shikudo code looks like it would work fine, but I'm not shikudo and can't test that.  I tried looking in the new documentation, but I didn't see anything related to the limbcounter.

VinceresEbonheart:
Tekura AXK still needs to  be updated to use lc_hit instead of ml_hit

PR #630, opened 2020-08-16
Limbcounter Update

K1ngSpad3z:
Updated earthLord limbcounter
  - Set limb break to static 5
  - Fixed a trigger

Updated Tekura limbcounter
  -Created configs for arm and leg damage under vconfig2
  -Updated triggers for new variables

Update Sylvan limbcounter
  - Moved it from the Meta limbcounter folder


K1ngSpad3z:
Resolves #625 
Resolves #622 
Resolves #610

PR #631, opened 2020-08-17
Updated Queue System

### LINKED ISSUE #629: Serverside Alises are now capped at 10 commands
Achaea pushed an update recently that capped serverside aliases to 10 commands instead of the original 20, as seen in Announce 5250. They also updated the CURING PRIORITY and CURING DEFENCE PRIORITY commands to allow multiple changes to each number

This will mainly effect curing priority changes that the system sets on re initialization through the alias it creates.

With these changes the serverside alias can be changed from:

`curing priority defence fangbarrier reset/curing priority defence third eyereset`

to:

`curing priority defence fangbarrier reset thirdeye reset`

Allowing us to send all the changes in afflictions and defences in a single command

https://cdn.discordapp.com/attachments/578659570973409355/744630083167322222/unknown.png

https://cdn.discordapp.com/attachments/578659570973409355/744630339682435173/unknown.png

K1ngSpad3z:
Creates a new function
  -sendcureingprio()
    -Used to store priority chanegs into a table that then sends to the doqueue as a single iteration instead of broken up

Also fixed the Inra alias
  -uses the in game alias now instead of a complex system (which caused alias command issues)

Resolves #629

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -2266,28 +2266,28 @@ end
 function svo.sendcuringprio(prio, iter, type)
   prio = " "..prio
   if type == 'affliction' then
-    table.insert(svo.sk.curequeue, prio)
-    if #sk.curequeue &gt;= 20 then
-      if #sk.sendqueue &gt;= 6 then
+    table.insert(svo.sk.affqueue, prio)
+    if #sk.affqueue >= 20 then
+      if #sk.sendqueue >= 6 then
         sk.dosendqueue()
       end
-      sk.sendqueue[#sk.sendqueue+1] = "curing priority"..table.concat(sk.curequeue)
+      sk.sendqueue[#sk.sendqueue+1] = "curing priority"..table.concat(svo.sk.affqueue)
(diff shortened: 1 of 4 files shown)


K1ngSpad3z:
Left a few things out that I thought I had put in there before hand, should be all good now

svo (curing skeleton, controllers, action system).xml
```diff
@@ -2264,6 +2263,35 @@ function svo.sendcuring(what)
...
+    if #sk.defqueue >= 20 then
+      if #sk.sendqueue >= 6 then
+        sk.dosendqueue()
+      end
+      sk.sendqueue[#sk.sendqueue+1] = "curing priority defence"..table.concat(svo.sk.defqueue)
+      if not sk.sendcuringtimer then
+        sk.sendcuringtimer = tempTimer(0, sk.dosendqueue)
+      end
+      sk.defqueue = {}
+    end
```
vadi2, line 2291: I'd put an else here in the end - so an error doesn't get swallowed if you typo affliction/defence.

K1ngSpad3z: Done

PR #632, opened 2020-08-18
Command Cap Fix

### LINKED ISSUE #629: Serverside Alises are now capped at 10 commands

K1ngSpad3z:
Creates a new function
-sendcureingprio()
-Used to store priority chanegs into a table that then sends to the doqueue as a single iteration instead of broken up

Also fixed the Inra alias
-uses the in game alias now instead of a complex system (which caused alias command issues)

Forces cap of 'multicmd' alias
-sends the doqueue before it can cap out at the command cap for Achaea

Resolves #629

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -1959,52 +1959,10 @@ signals.gmcpcharitemslist:connect(function ()
   if not gmcp.Char.Items.List.location then svo.debugf("(GMCP problem) location field is missing from Achaea's response.") return end
   if not sk.inring or gmcp.Char.Items.List.location ~= 'inv' then return end
 
-  local hadsomething = {}
-  for _, t in pairs(gmcp.Char.Items.List.items) do
-    if t.attrib and t.attrib:find('r', 1, true) then
-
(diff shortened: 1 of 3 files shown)


K1ngSpad3z:
Had to fix a few things between the last pull request and this one. Also added some code from Ada that helped fix a few issues. Should be cleaner, and in working order now

K1ngSpad3z:
I am going to go ahead and push this through only because we don't have a lot of testers right now through the developers and this will allow people to update and get a bigger group to report any errors based off of this change. I'll keep an eye on it and update any errors past this

PR #642, opened 2020-08-31
Properly recognize ablaze as a defence for Fire Lords

### LINKED ISSUE #638: Fire Lord Ablaze defence
ablaze is a defence for fire elementals, provides mild regen. It does not show up on DEF. This defence should be added somehow. 

When on keepup or defup and in the fire elemental class, the aff should be ignored. 

On: You flare your flames, burning ever brighter.
Off: Standard flame affliction cure lines
Continuous message while on: The fires crackling across your primal form bolster your eternal flame.
Syntax: MANIFEST TORCH
Already on: None, allows you to use it when already on.
Balanceful: yes, 2s

TheLastDarkthorne:
Ablaze actually regens firelord instead of hurting them, should be recognized as a defence.
This should work similarly to how blind/deaf switches works between affliction and defence according to special conditions,
Fixes #638

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -2333,6 +2333,19 @@ function svo.sk.fix_affs_and_defs()
     svo.addaffdict(svo.dict.deafaff)
     svo.echof("deafness is now considered an affliction, will cure it.")
   end
+  
+    --ablaze check for fire elementals, they always must be ON FIRE!
+  if affs.ablaze and ((defdefup[defs.mode].torch) or (conf.keepup and defkeepup[defs.mode].torch)
+    and (svo.me.class == 'fire Elemental Lord' or svo.me.class == 'fire Elemental Lady')) then
+    svo.rmaff('ablaze')
+    defences.got('torch')
+    svo.echof("ablaze is now considered a defence.")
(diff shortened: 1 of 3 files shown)


svo (curing skeleton, controllers, action system).xml
```diff
@@ -2333,6 +2333,19 @@ function svo.sk.fix_affs_and_defs()
...
   end
+  
+    --ablaze check for fire elementals, they always must be ON FIRE!
+  if affs.ablaze and ((defdefup[defs.mode].torch) or (conf.keepup and defkeepup[defs.mode].torch)
+    and (svo.me.class == 'fire Elemental Lord' or svo.me.class == 'fire Elemental Lady')) then
+    svo.rmaff('ablaze')
+    defences.got('torch')
+    svo.echof("ablaze is now considered a defence.")
+  elseif defc.torch and not ((defdefup[defs.mode].torch) or (conf.keepup and defkeepup[defs.mode].torch)
+   and (svo.me.class ~= 'fire Elemental Lord' or svo.me.class ~= 'fire Elemental Lady')) then
```
PhoenixCodes, line 2344: This should be an and, not an or, or it will always be a true statement. When you are a fire Lord, it is true that you are not a fire Lady

Issue #637, opened 2020-08-29
Sylvan Limbcounter


VinceresEbonheart:
svo limbcounter script wasn't updated to enable the sylvan triggers when you change to sylvan class when the recent changes were fixed.

Issue #596, opened 2020-02-16
Add pulverize lines to limbcounter


SlyViolin:
What I currently use to patch the gap.

^Throwing your full weight into the motion, you bring .+? warhammer across in a terrible blow toward (\w+)'s (.+)\.$

if svo.lc_checkdamage then
	svo.lc_hit(matches[2], matches[3], .5)
elseif not svo.lc_checkdamage then
	svo.lc_hit(matches[2], matches[3], 1)
end

PR #643, opened 2020-08-31
Limb counter change

awitteio:
Fixed Sylvan tracking

Added Pulverise trigger to Two Handed


K1ngSpad3z:
Fixes #637 
Fixes #596

VinceresEbonheart:
Earthlord Limbcounter still does not work the way it should. The way shaping works. You start at 0 shaping. You attack. Deal 1 hit damage. Increases shaping to 1. -- hit. Deals 1 damage. Increases to level 2. -- hit. Deals 1 damage. Increases to level 3. -- hit deals 1 damage. Increases to level 4. -- hit. Deals 2 damage. Stay at level 4.

The code is not checking your shaping level at the appropriate time. The way it currently coded -- is that it triggers UPON your attack. So you check shaping level, then it increases hit by 1 or 2. But by the time your attack is triggered, Achaea has already increased your shaping level, so by checking shaping level first, you're going to count hit inaccurately when you're at shaping level 3. ---- So If I'm at level 3 shaping. I attack -- triggered off my attack to check shaping (which  now shows as level 4 because I have attacked)...then it says I dealt 2 damage, which is not right...since my attack was initiated when I had 3 shaping and it should only deal 1 damage.

PR #644, opened 2020-09-01
Save Svo.me.path externally

DazedxJoker:


vadi2:
What does this do?

PhoenixCodes:
There were some issues with the limbcounter thanks to that path not being
loaded when the limbcounter loads. There may be a better way to handle
this, but... there it is.
----

DazedxJoker:
Svo.me.path is equal to "tekura" or "shikudo". However the variable wasn't saved like your class is. So, the limbcounter didn't know which form of monk to load.

vadi2 (approved):

PR #647, opened 2020-09-02
Tree cures

awitteio:
Fixed tree cures to properly track fractures

Fixed boosted Paradox from DW's


DazedxJoker (approved):
Treeing to cure fractures is working off gmcp now.

PR #657, opened 2020-09-29
Add gmcp balance tracking

DazedxJoker:
Also fixes some errors during Blackout


PhoenixCodes (approved):
This change creates a new bug, since the system goes up to the previous non-prompt line to write the balance timestamp. It should no longer do this, instead searching for the NEXT 'got balance' line to do this, or something else in blackout maybe. But. As this resolves issues with blackout that are very important, I'm voting for approving this, and fixing THAT issue down the road.

ndsamuelson (approved):
no issues that I was able to find. will miss you @PhoenixCodes .

Issue #624, opened 2020-06-28
Fishdist highlights not working


Twisteddoes:
Fishing highlights were not appearing in the config options and the alias to turn them on stopped as well. Dragonknight on the discord found the issue. The module priority in the scripts folder was set to 1 when it should have been 0. Changing this and restarting twice fixed it. 
![fishd1](https://user-images.githubusercontent.com/67557350/85955608-577e5300-b945-11ea-9eb3-0046e0cca7a9.png)

PhoenixCodes:
If this is the case that it being Priority 1 is breaking FishDistHighlights... Then there's something more wrong. FishDistPrio should not be loading at the same time as the system-level items - it is an add-on. Add-ons properly are supposed to load -after- the entire system; this prevents any functions they might call from not having been loaded yet. A priority of '0' is the highest priority, '1' is lower. 

This need to be looked into more fully and not just have a 0 priority slapped on the module and released.

DazedxJoker:
The issue is that svo system loaded event is being called before modules with priority 1+ is loaded.

PhoenixCodes:
Add-ons should not be dependant on that event then. They should assume the
system is already loaded if they are being installed/loaded.

Roland Goldberg
----

PhoenixCodes:
The reason I'm insistent on this is that it's difficult to control which things load first and which load last when they are at Priority 0. This can cause issues with various things that AREN'T controlled by the system loaded event, like the svo module list table.  I understand what you're saying about where the issue is, but that simply means the module needs to be corrected for the in-client setup, not have a kludge added that could break things down the road.

K1ngSpad3z:
Is this a case of creating a variable check for addons instead of relying on events to be loaded into the system?

PR #660, opened 2020-10-12
Allow fistdist addon to load on startup

DazedxJoker:


DazedxJoker:
Closes #624

vadi2 (approved):

PhoenixCodes:
Is FishDist properly using the loader then? I thought all the ones that were using the loader as intended were set to 0, and all the ones that weren't yet were set to 1.

DazedxJoker:
So, the svo system loaded event is called before addon's with a priority of . Any addon that uses that to load would need to have a priority of 0. It isn't actually incorporated into the loader. I believe this particular addon just adds a config on the event which is used to toggle the triggers.

PhoenixCodes (approved):

PR #663, opened 2020-10-20
Updates stupidCheck

PhoenixCodes:
Updates the commandCheck function to use rex.match, allowing stupidCheckTable to have regex values


DazedxJoker (approved):
Looks good

PR #666, opened 2020-10-21
Fix travis builds and possibly re-enable automated releases

keneanung:


precommit.lua
```diff
@@ -26,18 +23,3 @@ io.write(t)
...
 os.execute([[cd doc && sphinx-build -b html -d _build/doctrees . _build/html]])
-
--- update ndb cache listing
-package.path = package.path .. ";./doc/?.lua"
-gl     = require"parse_glossary"
-pretty = require"pl.pretty"
-
-local data = gl.striptrailinglines(gl.readfile("doc/namedb.rst"))
-
-local f = io.open([[svo template/ndb-help.lua]], "w+")
```
vadi2, line 37: Isn't this still needed?

keneanung: There are no templates anymore... But I could copy it to a helpfile right here :thinking:

keneanung:
We won't use travis anymore

PR #667, opened 2020-10-22
Waterlords doesn't need waterbubble

TheLastDarkthorne:
They are made of water you know, no need for bubbles here! Also avoid unnecessary pear/calcite eating while underwater for water lords.


DazedxJoker (approved):
Looks good

PR #668, opened 2020-10-22
More Mastery to my Weaponmastery!

### LINKED ISSUE #655: Mastery for Knight Classes
Some time ago (not exactly sure when) the Mastery skill was moved from Chivalry to Dual Cutting Weaponmastery skillset. This change needs to be reflected in the def table.

image

Needs to happen: Move Mastery defence to Weaponmastery for the dual cutting spec for Knights (Infernal, Paladin, Runewarden)

TheLastDarkthorne:
Mastery ability got changed to be in Weaponmastery for Dual Cutting specs. This should fix it and show the def only when these conditions for knight are met.
Fixes #655

### DIFF
--- a/svo (actions dictionary).xml
+++ b/svo (actions dictionary).xml
@@ -1062,11 +1062,14 @@ end
   },
   waterbubble = {
     gamename = 'airpocket',
+    onservereignore = function()
+      return svo.me.class == "water Elemental Lady" or svo.me.class == "water Elemental Lord"
+      end,
     herb = {
       def = true,
 
       isadvisable = function()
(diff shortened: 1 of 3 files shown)


PhoenixCodes (changes requested):
Something here is still breaking. Mastery is not adding when I use my newbie knight to test this. I think there need to be significant under-the-hood changes to how and when defs are added to the defence lists before we can do things like class specs separate from the class defs themselves. Would've been nice!

TheLastDarkthorne:
> Something here is still breaking. Mastery is not adding when I use my newbie knight to test this. I think there need to be significant under-the-hood changes to how and when defs are added to the defence lists before we can do things like class specs separate from the class defs themselves. Would've been nice!

Yeah... I will change it so it will show Mastery all the times now regardless of spec. At least it will be in the right place now, until we can figure out something with changing defs on the fly.

DazedxJoker (approved):
This works, until we can figure out a way to reload the defs for a specilisation change

PR #670, opened 2020-10-23
Complete revamp of the Inker module

Alyssea:
Same deal as before, still allows batching. No longer tries to cycle body parts, Achaea does that. Handles Maya card and stencils fine now.


PhoenixCodes:
Anaidianna would like to say she tested and it worked!

TheLastDarkthorne (approved):
It looks good, tested it with multiple tattoos on myself and on another friend. I couldn't test stencil though as I don't have the ability, but it should be ok too for the looks of it :) good stuff!

PR #672, opened 2020-10-26
Upstream Sync File

PhoenixCodes:
Execute this file from your local repo of your fork to sync your in-client branch to the upstream branch.

### DIFF
--- a/Sync_to_Upstream.sh
+++ b/Sync_to_Upstream.sh
@@ -1,0 +1,5 @@
+git checkout in-client-svof
+git fetch https://github.com/svof/svof.git in-client-svof
+git reset --hard FETCH_HEAD
+git push in-client-svof --force
+$shell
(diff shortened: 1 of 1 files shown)


Sync_to_Upstream.sh
```diff
@@ -0,0 +1,5 @@
+git checkout in-client-svof
+git fetch https://github.com/svof/svof.git in-client-svof
+git reset --hard FETCH_HEAD
+git push in-client-svof --force
```
keneanung, line 4: My bad. It needs a remote if given a branch name. But usually, you have a tracking branch configured anyways so this should be enough:
```suggestion
git push --force
```

Sync_to_Upstream.sh
```diff
@@ -0,0 +1,5 @@
...
+$shell
```
PhoenixCodes, line 5: I would also end with a newline, otherwise the $shell won't fire.

```sugestion
$shell

```

keneanung (approved):

Issue #530, opened 2019-10-02
Psion "super" slickness (Bloodflame) isn't registering


impuxi:
Psions have the ability to give a super slickness that can only be cured if you cure ALL bleeding on you and it auto cures once bleeding reaches 0. They need 250 bleeding to give it. Here is the log and the lines.

https://ada-young.com/pastebin/YLmcEVwZ
https://ada-young.com/pastebin/gywf6amA (this log I'm truelocked but system doesn't recognize that)

Diag Line: plagued with blood-fuelled flames.
Aff Delivered Line: The blood that flows from your wounds spontaneously bursts aflame, your skin bubbling and charring beneath its unnatural heat.

impuxi:
Another log: https://ada-young.com/pastebin/qrOO8vZA

Also seems to be some error at the end too.

impuxi:
Talked to DK about this issue a bit and it seems like it was already reported but I believe this issue has more details about it. Also, there doesn't seem to be a consensus on how to actually solve this which is pretty big since it is one of the main kill routes for Psions.

ndsamuelson:
#674 addresses this.

PhoenixCodes:
I believe #677 fixes this. I'd like someone to test and report please!

TheLastDarkthorne:
Very old issue that seems to not been fixed yet. But with gmcp affliction tracking being a thing now, shouldn't be hard to determine whether or not this is up now.
Implementing a proper logic to curing this though might be a bit tricky though. Something amongst the lines of if havethisaff and isbleeding and mana > manasipthreshold then clot clot clot until it is registered cured.

PR #677, opened 2020-10-28
added lightbind, no content yet

ndsamuelson:


PhoenixCodes:
Does this fix #530? Looks like it might. What is lightbind?

PR #678, opened 2020-10-29
Fixed an AI folder and added some common donext()'s

ndsamuelson:


TheLastDarkthorne:
You should probably make your fixes to the AI folder into a different PR alone. Cannot approve this while you have a trigger for lightbind without any content since it won't do anything. If you are still working on the lightbind stuff, consider turning that into a draft PR while you figure it out.

PhoenixCodes:
Those triggers were added for me, and was a PR goof that they ended up together. We still need a backend for psion power. If it works other than those empty trigs, I would say this gets an approval @TheLastDarkthorne

PhoenixCodes (approved):
Those donexts are good!

PR #679, opened 2020-10-30
Updated elistsorter

TheLastDarkthorne:
Counting things properly now, removed functions for pouring and disposing of older vials (consolidate vials is a thing now in achaea and no one carries thousands of vials anymore), no longer tracking and trying to refill unknowns, now tries to refill from rift as well.
Work in progress still need some fixes


PhoenixCodes:
Did not know about consolidate vials. Getting a dispose line might still be nice, so you don't have to manually give each vial number to a humgii or Rufio.

PR #681, opened 2020-11-03
Properly disable monk arm balances when not monk

DazedxJoker:
Case sensitivity wasn't allowing the Monk balances to disable when it should


PhoenixCodes (approved):

PhoenixCodes:
This may overwrite some previous PRs... I think we need to merge those before we do this, then fix this. I am going to assume that the bloodflame fix works, and get that merged. Please recommit afterwards with updated files

PR #714, opened 2021-01-27
Fix description of customprompt colortag

TheLastDarkthorne:
The ^6 colortag also colours deptshwalker age according to the amount, the help file should explain that it does so.


vadi2 (approved):

vadi2:
Hey thanks for the fix!

PR #727, opened 2021-07-21
Changes to Logging

awitteio:
Allows us to properly turn the logger off without having to reset the system or mudlet


vadi2:
Mm how was this breaking things?

awitteio:
> 
>
> Mm how was this breaking things?

When the system first starts, the variable for conf.log is set to "off" but turning the logger on sends this function, and the if statement for "off" has it set the variable to false. The last else in this if statement on the function calls the logger because the variable is set to false and not "off". So removing the part where it sets the variable to false, keeps the logger set to "off" when set initially. Meaning we don't have to re-initialize the system or restart mudlet to get the variable to set back correctly.

vadi2 (approved):

PR #733, opened 2021-07-22
Hypochondria fixes

awitteio:
Changes pool of afflictions to track for Hypochondria

Changes time to remove counts for hypochondria from 12 to 3 to account for hypochondria affliction tick rate

### DIFF
--- a/svo (actions dictionary).xml
+++ b/svo (actions dictionary).xml
@@ -6894,11 +6894,6 @@ end
         else
           svo.affsp.anorexia = true
         end
-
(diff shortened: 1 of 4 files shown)


awitteio:
This is a more updated version. Only tracks Hypochondria from the 4 affliction ticks every 2 seconds and only assuming that you don't have Hypochondria already.

TheLastDarkthorne:
It looks good to me and worked as well on my little tests although I couldn't test it extensively... The code seems fine too so I don't have anything against merging it!

PR #748, opened 2021-11-24
Streamline addons to use svo.loader

TheLastDarkthorne:
Lotsa addons not using the proper svo loader and instead doing their own thing! Now they will sit comfortably and wait the loader to call them like the good little addons that they are!


svo (refiller).xml
```diff
@@ -542,14 +541,15 @@ function svo.rf_missingstuff()
   echo'\n' svo.echof("Ack, looks like you're out of enough ingredients. :(")
   svo.rf_cancel()
 end
-
-end)</script>
+print("\nREFILLER LOADED !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
```
vadi2: debug?

TheLastDarkthorne: oop lol I will remove that later

Issue #686, opened 2020-11-12
remove riding as a keepup in elemental lord, as well as blocking.


ndsamuelson:
since elemental lords cannot mount steeds or block **(I know at least waterlord can't, not sure on the others)** this ends up looping over and over if you add to keepup accidentally or whatever the case may be.

QwindorRousseau:
Pretty sure you can block as earth lord but i need to verify when i get in game again

ndsamuelson:
yeah I mean you can hide the defs I suppose, I just mean as far as polish goes for the new user and QoL overall. Nothing ***earth-shattering*** heh.  Just checking things off my list:
- [x] Mention all the things
- [x] Nail it.
- [ ] Break more things.

QwindorRousseau:
Confirming that Earth Lord does have the ability to block for reference on whoever is working on this.

PR #753, opened 2022-01-14
Riding defense Logic

tsartin:
Updated the riding defense logic to not spam Elemental Lords/Ladies, since they cannot use a mount. This should solve issue #686


vadi2 (approved):

TheLastDarkthorne:
Looks good to me, I can't test it though cause I don't have riding but I should work fine that way. I will wait a few days to see if the other devs have a saying before merging.

Issue #695, opened 2020-11-18
Tattoo Inker does not accept a limb argument 


sdelo1185:
Does not work when trying to ink tattoos. Have to turn off the tattooing alias and manually tattoo.

PhoenixCodes:
Can you show a log of this? The tattooer was recently overhauled. As far as I know, it only breaks if someone else in the room has a failed tattoo ink at the same time, and that's due to Achaea not showing which tattooist has failed to ink.

sdelo1185:
https://ada-young.com/pastebin/kkybjhjX

K1ngSpad3z:
update to the latest version of the in client system and try it again, I have tried to figure out a way to replicate this error and can't. Its working as intended on my end. (in game do vupdate force to get the latest build for the system)

sdelo1185:
Seems that when i use the new ink alias it works fine if i just do ink (tattoo name)

PhoenixCodes:
oh, I see what happened. You selected a body part. Inker does not support that, but should. I'll mark this as an enhancement request.

sdelo1185:
Does this accept an argument for inking on others?

tsartin:
This can probably be closed.

ndsamuelson:
> This can probably be closed.

Can confirm.

PR #754, opened 2022-01-16
Inker Changes

tsartin:
1. allows limb arguments
2. adds auto reinking toggle for interrupted tattoos

### DIFF
--- a/doc/inker.rst
+++ b/doc/inker.rst
@@ -1,24 +1,30 @@
 Svof Tattoo Inker
 =================
-This script simplifies the process of inking tattoos! Does your tattoos, and tattoos for other people - and allows queueing up of several tattoo orders at once, so you can do a newbies set all with one command, for example. It outrifts inks, the body for a free location, and inks the tattoo. Also touches tattoos requiring activation, or if its for someone else, tells them which tattoos to touch at the end. Does ink on the back bodypart as well.
+This script simplifies the process of inking tattoos! Does your tattoos, and tattoos for other people - and allows queueing up of several tattoo orders at once, so you can do a newbies set all with one command, for example. It outrifts inks, the body for a free location, and inks the tattoo. Also touches tattoos requiring activation, or if its for someone else, tells them which tattoos to touch at the end.
 
 If inking gets interrupted, it'll show a warning, wait a bit, and resume inking again.
 
-You can also use *vconfig telltouch* to tell people to touch the tattoos that need touching after inking.
 
+Configs
+^^^^^^^^
+**vconfig telltouch** 
+  tell people to touch the tattoos that need touching after inking.
+
+**vconfig autorink**
+  toggle whether or not you auto reink a tattoo after being interrupted
+  
 Aliases
 ^^^^^^^^
-.. glossary::
+**ink <tattoos> on <person>**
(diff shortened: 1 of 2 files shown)


tsartin:
1) Allows limb arguments in inker as requested in #695
Note that it works for the syntaxes:

ink boar 
ink boar on person 
ink boar on limb of person

But does NOT work for the syntax:

ink boar on left arm (which normally does work on yourself in game)

I'm not sure how to add that and not conflict with 'ink boar on person'.

If accepted the documentation should also be updated. (I'm not sure how to change the documentation)

tsartin:
2) Default was to always reink interrupted tattoos which could be frustrating.

This just adds an option to turn that off.

The proposed option would be 'vconfig autoreink'

vadi2 (approved):

TheLastDarkthorne:
> 1. Allows limb arguments in inker as requested in [Tattoo Inker does not accept a limb argument  #695](https://github.com/svof/svof/issues/695)
>    Note that it works for the syntaxes:
> 
> ink boar ink boar on person ink boar on limb of person
> 
> But does NOT work for the syntax:
> 
> ink boar on left arm (which normally does work on yourself in game)
> 
> I'm not sure how to add that and not conflict with 'ink boar on person'.
> 
> If accepted the documentation should also be updated. (I'm not sure how to change the documentation)

You can probably make a table with all acceptable limb parts and check the match against it. If it is not a limb part, assume it is a person then?

TheLastDarkthorne:
Also to edit the docs you just go in the docs folder and commit an edit to the respective doc files in your branch

tsartin:
> > 1. Allows limb arguments in inker as requested in [Tattoo Inker does not accept a limb argument  #695](https://github.com/svof/svof/issues/695)
> >    Note that it works for the syntaxes:
> > 
> > ink boar ink boar on person ink boar on limb of person
> > But does NOT work for the syntax:
> > ink boar on left arm (which normally does work on yourself in game)
> > I'm not sure how to add that and not conflict with 'ink boar on person'.
> > If accepted the documentation should also be updated. (I'm not sure how to change the documentation)
> 
> You can probably make a table with all acceptable limb parts and check the match against it. If it is not a limb part, assume it is a person then?

That's a good idea. I had actually thought about something similar then I got stuck on the actual match. Currently it looks for "(%w+)$" but it needs to match ' one or two words then end of line'  🤔

TheLastDarkthorne:
@tsartin `local anotherperson = order:match(" on (%w+)$") or order:match(" of (%w+)$")`
The problem with this is that it will also match 'ink boar of person' for example, not that anyone would deliberately do that but I don't think it is a proper way to handle the problem... 
You can probably change the regex to account for multiple words (like .+)  and then check it against a table of bodyparts?

EDIT: Actually I think there might be a better way to handle this which is to change the alias regex to something less greedy... I will see if I can come up with something

TheLastDarkthorne:
Could try changing the alias regex to this, lookie:
https://regexr.com/6diud

This way, you only need to handle the matches instead of the whole string, much better! (I think)

tsartin:
> a better way to handle this which is to change the alias regex to something less greedy... I will see if I can come up with something

I think I prefer this too.  If someone does 'ink boar' there's not a whole lot the inker adds over the in-game inker anyway. 

Then do we just do check on group 2? Like if group two is not a limb then it's a person?

TheLastDarkthorne:
> > a better way to handle this which is to change the alias regex to something less greedy... I will see if I can come up with something
> 
> I think I prefer this too. If someone does 'ink boar' there's not a whole lot the inker adds over the in-game inker anyway.
> 
> Then do we just do check on group 2? Like if group two is not a limb then it's a person?

Yup! :)

Issue #693, opened 2020-11-18
Svo.classes not flagging true for Psion, Elemental Lords.


sdelo1185:
The auto function for flagging these classes does not turn on the classes for Psion or the Elementals. Psion can be manually turned on and off via TN PSION TF PSION, but there's currently no way to turn on the elemental lords because the command doesn't recognize that they exist.

sdelo1185:
Changing the Table in svo Class Management to

svo.classes = {
  airelemental = {fighting = false, skills = {"duress"}},
  alchemist    = {fighting = false, skills = {"transmutation", "physiology", "alchemy"}},
  apostate     = {fighting = false, skills = {"evileye", "necromancy", "apostasy"}},
  bard         = {fighting = false, skills = {"voicecraft", "swashbuckling", "harmonics"}},
  blademaster  = {fighting = false, skills = {"twoarts", "striking", "shindo"}},
  depthswalker = {fighting = false, skills = {"shadowmancy", "aeonics", "terminus"}},
  dragon       = {fighting = false, skills = {"dragoncraft"}},
  druid        = {fighting = false, skills = {"groves", "metamorphosis", "concoctions"}},
  earthelemental = {fighting = false, skills = {"sculpting"}},
  fireelemental = {fighting = false, skills = {"ignition"}},
  infernal     = {fighting = false, skills = {"oppression", "chivalry", "weaponmastery"}},
  jester       = {fighting = false, skills = {"tarot", "pranks", "puppetry"}},
  magi         = {fighting = false, skills = {"elementalism", "crystalism", "enchantment"}},
  monk         = {fighting = false, skills = {"tekura", "kaido", "telepathy"}},
  occultist    = {fighting = false, skills = {"occultism", "tarot", "domination"}},
  paladin      = {fighting = false, skills = {"chivalry", "devotion", "weaponmastery"}},
  priest       = {fighting = false, skills = {"spirituality", "devotion", "zeal"}},
  psion        = {fighting = false, skills = {"weaving", "psionics", "emulation"}},
  runewarden   = {fighting = false, skills = {"runelore", "chivalry", "weaponmastery"}},
  sentinel     = {fighting = false, skills = {"metamorphosis", "woodlore", "concoctions"}},
  serpent      = {fighting = false, skills = {"subterfuge", "venom", "hypnosis"}},
  shaman       = {fighting = false, skills = {"spiritlore", "curses", "vodun"}},
[...]

PR #756, opened 2022-01-28
Class detection changes

tsartin:
Changes elemental lord format in svo.classes. Changed "Water Elemental" to "Water", etc

Also added some triggers to help detect elemental lords and Psion.

Should allow class flagging to work via triggers and TN/TF for elemental lords and psion as requested in #693

I changed this according to the suggestions but did not use the 'svo.classchange' suggestion as that refers to personal class and not who you are fighting.

NOTE: One concern is if someone was using something like svo.enableclasses or svo.classes with the old class names. Since there were no triggers to detect those classes, this it may not be a big deal though


TheLastDarkthorne:
Have you tested if the classes are flagging properly with this?

tsartin:
I just double-checked to be sure. The classes are flagging correctly when you fire svo.startedfighting() and svo.valid.gothit().

Then they show true in svo.enabledclasses. 

Also seems to be disabling correctly as well. 

I did a quick scan of the new triggers for typos as well and didn't notice anything but that seems to be the only fail point for not flagging correctly.

PR #758, opened 2022-02-01
Change Aria defense to cure deaf if up

tsartin:


vadi2 (approved):

svo (actions dictionary).xml
```diff
@@ -12056,7 +12056,10 @@ if svo.haveskillset('voicecraft') then
       end,
 
       action = "sing aria at me",
-      onstart = function() send('sing aria at me', conf.commandecho) end
+      onstart = function()
+        if defc.deaf then send('apply epidermal to ears', conf.commandecho) end
```
TheLastDarkthorne: I'm not sure this is going to work though? Cause the `isadvisable` up there is set to not put this defence up if the player is deaf so it will never run in that case. Could perhaps remove that section of the isadvisable if you are sure that bards wants to keep this defence up even if they are deaf.

tsartin: Good catch. Just double-checked with a bard. That seems to be the goal of this change. They want to be able to use aria while still being flagged as deaf. I miss understood the assignment. I'll probably replace the deaf check with a salve balance check and maybe a slickness check.

TheLastDarkthorne: Sounds okay ^-^ just be sure it won't block upon slickness alone as that will not impede the defup of aria in all situations, only when they can't apply the salve.

TheLastDarkthorne: I was thinking on this... in the onstart() function should make it so that if are deaf send the command with an apply epidermal if not def then just send the normal command or else it will always apply the salve even if they are not deaf.

tsartin: Sorry for the absence. That seems much better. Should there be a check for slickness in there? Or is that overkill for this?

TheLastDarkthorne: I think it is probably overkill, but if you want to test in case you have slickness and try to put aria up and see how it behaves it would be nice. Otherwise, just give me a headsup if you want to work on this still or we can merge it in the state it is now.

tsartin:
@TheLastDarkthorne I think we can just leave it for now. I don't have as easy access to it at the moment. If someone has an issue with it we add that in later.

vadi2:
@TheLastDarkthorne ?

TheLastDarkthorne:
This sounds good, sorry for taking forever to merge this, end year, start year, holidays and stuff making my life crazy lol

Issue #761, opened 2022-02-26
Instakill warning for Zeal Judgement ability


TheLastDarkthorne:
Judgement still has no instakill warning, it should be added sometime.
It looks like:

```
Person clasps his hands in solemn supplication, a recital of your many sins spilling forth in a fervent prayer.

The prayer of Person builds in strength as he fixates upon you, righteous fury evident in every word.

As the righteous words of Person build towards an ominous conclusion a light begins to shine upon you, its merciless glare a herald of danger.

A terrible flash and an instant of soul-searing agony heralds the conclusion of Person's prayer, the merciless glare transforming to a cataclysmic blast of effulgent fury.
```

tsartin:
Don't you do it stale bot

PR #762, opened 2022-03-28
Instakill warning for Zeal Judgement ability

tsartin:
Instakill warning for Zeal Judgement ability #761

I added the last trigger listed but I'm not sure if that one is death or the last warning.

### DIFF
--- a/svo (aliases, triggers).xml
+++ b/svo (aliases, triggers).xml
@@ -42510,6 +42510,109 @@ svo.valid.simpleunknownany()</script>
 						<integer>1</integer>
 					</regexCodePropertyList>
 				</Trigger>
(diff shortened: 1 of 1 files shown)


TheLastDarkthorne:
Hi! I'm sorry I have not looked into your PRs lately. Just too much going on here
Anyway, the last line is the death line afaik so warning on it is not a good thing (can always confirm it first tho if not sure) ! It should warn when it starts and perhaps have a few more as it continues (similar to devour and others)

tsartin:
@TheLastDarkthorne I made the last line a trigger here but it doesn't actually do anything. Only the first 3 triggers have a warning!

PR #776, opened 2022-08-12
update gmcp balance tracking

tsartin:
This will remove tracking balance and equilibrium from the prompt and solely use gmcp. This should help with things like tracking balance in blackout which had some issues and also remove some odd fixes throughout svof. 

There seem to be a few oddities tracking eq/bal in blackout pre-gmcp. I tried to catch them all and remove them without breaking something else but we will see. No issues that I could tell so far.

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -2534,8 +2534,8 @@ function svo.valid.setup_prompt()
     bals.rightarm = 'unset'
     bals.leftarm = 'unset'
   else
-    bals.balance = false
-    bals.equilibrium = false
+    --bals.balance = false
+    --bals.equilibrium = false
     svo.pflags = {}
   end
 end
@@ -2649,6 +2649,18 @@ function svo.onprompt()
   sk.systemscommands = {}
 
   svo.promptcount = svo.promptcount + 1
+       
+  if svo.conf.showbaltimes then 
+    
+    if svo.sk.balchange and svo.stats['lastbalance'] then
+      svo.echotime(svo.stats['lastbalance'])
+      svo.sk.balchange = nil
+    end
+    if svo.sk.eqchange then
+      svo.echotime(svo.stats['lastequilibrium']) 
+      svo.sk.eqchange = nil
+    end
+  end
 
   checkblackout()
   check_promptflags()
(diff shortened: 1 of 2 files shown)


svo (curing skeleton, controllers, action system).xml
```diff
@@ -3026,6 +3040,7 @@ end
 svo.savesettings = svo.QQ
 
 -- add in blackout only, otherwise go off the prompt - this allows for time tracking
+--maybe can remove this now if it was only for tracking eq in blackout?
```
vadi2: This was to protect against illusions that would make you think that you are off balance, or disrupt in blackout where you'd come out with no eq and would never regain it. If it's not a thing anymore in the game - then go for it

tsartin:
@vadi2 Illusions shouldn't matter much anymore but disrupt in blackout does still behave a little odd. I think we can keep this for now. It could probably be rewritten to account for GMCP but this should still work for now.

Basically when in a blackout, if you get disrupted you keep EQ via gmcp but as soon as the blackout ends gmcp should notice but there may be other situations I'm not sure of.

vadi2:
Sounds good. Could you update the comment?

vadi2 (approved):

TheLastDarkthorne:
@tsartin have you tested and confirmed that this worked? Sorry for taking so long to come to this one. If it is then I will include it in the next update already

tsartin:
@TheLastDarkthorne I vaguely remember it working at the time but it has been awhile if you'd like me to review it one more time. 

I'm away from my computer this weekend but can check it early next week.

tsartin:
This seems to track well however....I'm getting a weird space when i gain balance/eq so need to figure that out first.

image

TheLastDarkthorne:
> This seems to track well however....I'm getting a weird space when i gain balance/eq so need to figure that out first.
> 
> image

Did you figured if this was caused by the change or just some mudlet kirk?

tsartin:
@TheLastDarkthorne sorry for the delay. I loaded this up and it doesn't seem to be having this issue so was probably something weird back then


image

TheLastDarkthorne:
> @TheLastDarkthorne sorry for the delay. I loaded this up again and it doesn't seem to be having this issue so was probably something weird back then
> 
> image

Weird I loaded it up and it didn't do that for me either lol 
Anyway! Since we are firing this off gmcp now, we probably don't need the 'svo got balance' and 'svo got eq' triggers anymore so maybe should delete those? Those are probably not necessary anymore I guess (or only activate it during blackout maybe since gmcp info is not available on blackout? Not sure if blackout affects balance gmcp anyway)

PR #780, opened 2022-10-05
Update triggers to prepare for Fae/Faen pronouns

AlyxRen:
Based on Phaestus' suggestion, I've done the work of converting pronoun based triggers that are tracking other adventurers to use \w+ instead of the myriad of options used to cover the various pronouns. This was suggested in the Achaea Discord, to be future proofed against other potential neopronouns being added down the road.

I've tested the NameDB changes in particular, the Combat ones are a little harder to test. Planning to test them later today when I can, and update this PR with the results. But I wanted to get this out here for review and comments.


TheLastDarkthorne:
Hello! Thank you for wanting to contribute to the project :)

I'm not sure if it is a good idea to replace everything with greedy regex like this. I mean, it works if you are expecting random words, but we know what the pronouns will be, it is just a matter of adding another '|'.
Will wait to know what @vadi2 thinks about this tho, if he says it is okay then fine

keneanung:
Vadi is on a well deserved vacation, so this might sit a while if we wait.

While I usually agree to be as specific as possible, the performance hit here is usually negligible up to non-existent. The start of the line has to match already, which makes it unlikely that the rest of the line including the pronoun does not match.

The only exception is when the match is at the start of the trigger, which might cause a few unnecessary steps. Maybe a before/after comparison with the stressinator might shed some light?

AlyxRen:
The reason I did it like this, instead of updating the hodgepodge of specific regexes (There was `him|her` and `her|him` back to back. `s?he` is performant, but a nightmare to read, and there was even a `him|him` bug that gets swept up and fixed here) is because Phaestus specifically called out more greedy wildcarding to avoid having to do this in the future.

https://discord.com/channels/872599744793419776/874108288957034556/1026267048235511950
> I suggest you take this opportunity to replace them with wildcards instead of hardcoded pronouns lists to avoid having to repeat this in the future

AlyxRen:
If you come to the conclusion as the regular maintainers of svof that you'd rather ignore that warning, and hard-code them, I can redo the work, and make it more consistent so that future problems are somewhat easier to update with a few dedicated Find and Replace's.
The major issue with this method, is that while a little more performant, it requires constant vigilance to not fall into the same traps of using different trigger forms that achieve the same goal in the future, like the `him|her` vs `her|him` issue. And the more pronouns that end up added to that list, the more likely to one of the lines to end up in a different order, potentially getting missed the next time you have to change anything.
They probably aren't adding dozens of options, but with the admin specifically calling out future changes, I think  another couple could be likely.

TheLastDarkthorne:
> If you come to the conclusion as the regular maintainers of svof that you'd rather ignore that warning, and hard-code them, I can redo the work, and make it more consistent so that future problems are somewhat easier to update with a few dedicated Find and Replace's. The major issue with this method, is that while a little more performant, it requires constant vigilance to not fall into the same traps of using different trigger forms that achieve the same goal in the future, like the `him|her` vs `her|him` issue. And the more pronouns that end up added to that list, the more likely to one of the lines to end up in a different order, potentially getting missed the next time you have to change anything. They probably aren't adding dozens of options, but with the admin specifically calling out future changes, I think another couple could be likely.

It is all good. If you can, see if you are able to find some people to test some of the combat triggers to make sure they are working.

AlyxRen:
I'm working to find the other lines and get them tested, but I have a tritrans jester, so I did test with the Devil Tarot defense, and they are all working, barring the above hangedman bug that was looking for the target's name to be all numbers.

PR #781, opened 2022-10-08
Fix ship prompt disable trigger

TheLastDarkthorne:
Probably related to deleteLine not deleting prompts properly sometimes.


TheLastDarkthorne (commented):
-

PR #786, opened 2022-10-19
Fixes being unable to ink on a specific leg

codeacula:
The regex as it is looks for either a space followed by "arm" or just "leg". This means that with SVO right now we can't target left/right legs. Confirmed this is an SVO issue by adding spaces to the command "ink tree on right leg of Person" and it worked, but without them svo said "(svof): Going to ink tree on open place of right."

### DIFF
--- a/svo (inker).xml
+++ b/svo (inker).xml
@@ -183,7 +183,7 @@
 			<script>svo.ti_ink(matches[2], matches[3], matches[4])</script>
 			<command></command>
 			<packageName></packageName>
-			<regex>^ink (.+) on (\w+(?: arm| leg)?)(?: of (\w+))?</regex>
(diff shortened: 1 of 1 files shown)


TheLastDarkthorne:
Nice catch there, I really overlooked it. 
Also taking a look at this regex, maybe the first capture should be (\w+) instead of (.+) as I think the latter captures multiple words but the tattoo names all have only one word?

codeacula:
Changed

tsartin:
The first (.+) here is because it is made to queue multiple tattoos.

For example 'ink moon moss tree on john'

TheLastDarkthorne:
> The first (.+) here is because it is made to queue multiple tattoos.
> 
> For example 'ink moon moss tree on john'

Ahh right... sorry I forgot about that

codeacula:
Reverted

PR #788, opened 2022-10-24
deleted duplicated .xml

ndsamuelson:
unsure if this is the one that's unwanted, so feel free to delete the other. But either way you're duplicating the scripts in mudlet.


ndsamuelson:
disregard, the duplicate was purposeful, according to Erikka. Closing for now.

PR #791, opened 2022-10-31
Add support for releases again - master

TheLastDarkthorne:
Hello!
This is the first of a series of PRs I'm gonna make (since it also has changed on a few other branches for it to work) to add support for proper releases for Svof, as well as a (admittedly very basic) versioning logic.

The goal for this is to only create a release when we are 100% ready for it and not when pull requests are merged. So I made it so it will trigger the chain of workflows once the file 'current_version.txt' in gh-pages is edited with a new version which signifies we are ready for a new release.

For this to work though, we will need to create a new PAT to be used in one of the workflows. This is so because github can properly chain those workflows (if use `secrets.GITHUB_TOKEN` alone it won't chain them).

I took the liberty to create and edit a few labels too, since we will need the labels so the changelog builder can run.

With this we can probably finally close up the Client-Side Release project.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,26 +21,13 @@ jobs:
           # specify here the path to your version file (e.g. package.json, pom.xml...)
           version-file: inclient/current_version.txt
           schema: major
-          
-      - name: Verify tag
-        id: verifytag
-        env:
-          VersionToVerify: ${{ steps.getversion.outputs.version }}
-        run: |
-          echo "$VersionToVerify"
-          if [ $(git tag -l $VersionToVerify) ]; then
-            echo "{pushtag}={false}" >> $GITHUB_OUTPUT
-          else
-            echo "{pushtag}={true}" >> $GITHUB_OUTPUT
-          fi
         
     outputs:
       version: ${{ steps.getversion.outputs.version }}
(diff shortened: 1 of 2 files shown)


vadi2 (approved):
Looks good! Could you document how to do a new release with this method?

TheLastDarkthorne:
> Looks good! Could you document how to do a new release with this method?

I can put it in the developer readme when we are done, But it is very straightforward, merge pull requests and when it is ready for a new release just raise the version in the `current_version` file and it will fire the release process. I thought this method would be safer since we already have to edit the file anyway (since svof checks for it client-side) and it is pobably safer to do it this way or with workflow_dispatch trigger than on push (not always we want to release a new version for every merged PR and just bunch them all together in a new release for example)

TheLastDarkthorne:
Closing this since the workflows will go to the in-client branch which will become the new supreme master branch!

vadi2:
Oh, why? Maybe we had a miscommunication - but we should keep `master` as the main branch per convention*.

* or rename it to `main`

TheLastDarkthorne:
Oh I mean, since the inclient branch is going to be master branch when we finish so putting the workflows here is useless anyway, I just moved the other work flow to the inclient branch and edited them to refer to master

PR #793, opened 2022-10-31
Add support for releases again - in-client-svof

TheLastDarkthorne:
Hello!
This is the first of a series of PRs I'm gonna make (since it also has changed on a few other branches for it to work) to add support for proper releases for Svof, as well as a (admittedly very basic) versioning logic.

The goal for this is to only create a release when we are 100% ready for it and not when pull requests are merged. So I made it so it will trigger the chain of workflows once the file 'current_version.txt' in gh-pages is edited with a new version which signifies we are ready for a new release.

For this to work though, we will need to create a new PAT to be used in one of the workflows. This is so because github can properly chain those workflows (if use `secrets.GITHUB_TOKEN` alone it won't chain them).

I took the liberty to create and edit a few labels too, since we will need the labels so the changelog builder can run.

With this we can probably finally close up the Client-Side Release project.

Example release:
https://github.com/TheLastDarkthorne/svof/releases/tag/untagged-1ae1014320353b75f62e


vadi2:
We should only merge it once and then update other branches with this code. For example the in-client-svof branch will go away, right?

TheLastDarkthorne:
> We should only merge it once and then update other branches with this code. For example the in-client-svof branch will go away, right?

Oh... yeah, my bad lol 
The master one is prob not needed indeed, just add it all here. I will do it soon

TheLastDarkthorne:
Okay! I just need to probably copy this into my repo's master and see if it works there but it should be good to go now.
Would be nice if you could take a look at all the changes and see if it is all okay to you. I made a few changes about the release channels that are important:
1 - The main releasechannel now is 'stable'
2 - There is also a 'testing' release channel that is going to download the xmls directly from the repo instead from the release (just as it is doing now with no releases).
3 - For this to work, I also made a few changes to the gh-pages which include getting rid of all the old releases that were stored there since we are not using them anyway. Check the gh-pages PR to see the changes.
4 - I kept the inclient folder in gh-pages so that older versions doesn't break. Probably should keep the in-client-svof branch as a backup for these older versions but not update it anymore. Or we can just move on and tell people to download the new release version manually.

I thought this is a good way to handle it because we can ask people to test a release before it actually goes out or we can merge PRs in the master branch without it automatically triggering a new release, in which we can control when we actually want to make new releases while keeping an avenue for people to test new changes or a batch of new changes.
But of course, I'm not a professional so I want to ask if it you think it is good, if not we can just keep only one release channel or discuss other possibilities.

TheLastDarkthorne:
Also if you approve this, I think only you can actually move branches around and rename them or set them to master and stuff cause I'm not admin. Remember we need that new PAT as well for the workflow chain to go too!

vadi2:
> 1 - The main releasechannel now is 'stable'

Devs like to call things stable but users always expect them to be stable... is there a better word to use here than that's non-techy?

> 2 - There is also a 'testing' release channel that is going to download the xmls directly from the repo instead from the release (just as it is doing now with no releases).

Sounds good

vadi2:
Lets rename the stable channel and then do the move and add PATs

TheLastDarkthorne:
> Lets rename the stable channel and then do the move and add PATs

Hmm, could rename it to 'main' maybe? Not sure what name would be best

vadi2:
That sounds like a good one 👍

PR #796, opened 2022-11-10
Emergency fix for Mudlet's sync bug wiping out files

TheLastDarkthorne:
Due to a bug in Mudlet's module sync, need to disable sync first or it will wipe out the xmls. Putting this bandaid fix so people don't run into it anymore until it is fixed. Also fixed an unrelated problem with the naming of one of the modules in the list.


keldin-coding:
Just commenting with the Mudlet issue to link them so if anyone comes looking they can see and follow along or see what was done here: https://github.com/Mudlet/Mudlet/issues/6414

Issue #779, opened 2022-09-30
Monk->Bard Class swap


Embrelle:
Swapping from Monk to Bard causes "[string \"Script: Setup\"]:431: attempt to index field 'tekura' (a nil value)" to appear on every line. A makeshift fix that's worked for me has been to change line 135 in the setup script from:
svo.sk               = svo.sk or {}
to
svo.sk               =  {}
which has not caused any further issues thus far.

tsartin:
@Embrelle Are you still having this issue? And if so, does it do this on a new profile?

Embrelle:
@tsartin Yes, just tested with a fresh profile and the problem is still there. As a note for replication I've tried this with both Shikudo and Tekura monk, and the error message disappears after restarting mudlet but will come back if I switch to monk and then to bard.

tsartin:
@Embrelle could you post a screen shot of what it looks like right after you change your class to bard? I'm trying to figure out where the switch is failing. 

It should have something like 
"(svof): Welcome back!
(svof): Loaded & ready to go, version 52 (homepage)
...etc"

Embrelle:
![image](https://user-images.githubusercontent.com/114719949/200421730-53fb05b7-0f1e-416b-b645-c79870e01360.png)
@tsartin

Embrelle:
@tsartin your proposed change, to add svo.signals.gmcpcharvitals = nil to svo.classchange resolved the problem!

PR #797, opened 2022-11-14
Reset all signals on init

tsartin:
This helps overlapping signals from causing problems after a class change ( or anytime the system reinitializes)

Should solve #779

also solves another error that svof seems to be throwing from the class change script. Not sure when this started but it is in the main build that is live.

image


tsartin:
Side note, I did test this with class changes but would love if someone else could as well.

vadi2 (approved):
Let's try it

TheLastDarkthorne:
I did a few tests in here and it seems ok. Hopefully it won't break stuff for other people lol

PR #806, opened 2023-01-11
Fix release 54

keneanung:


svo (install me in module manager).xml
```diff
@@ -7187,8 +7190,7 @@ end</script>
 				<name>svo.classchange</name>
 				<packageName></packageName>
 				<script>function svo.classchange(override)
-
-  local newclass = (override ~= "gmcp.Char.Status") and override or gmcp.Char.Status.class
+  local newclass = (override ~= gmcp.Char.Status.class) and override or gmcp.Char.Status.class
```
TheLastDarkthorne: Commenting here so I know where to look later, this line does not need changing and actually it is breaking the whole classchange thing making it always set to "None"

svo (trigger functions).xml
```diff
@@ -4624,70 +4624,83 @@ end
...
+    svo.dict.gotbalance.happened.tempmap[#svo.dict.gotbalance.happened.tempmap+1] = 'word' -- hack to allow multiple balances at once
+    lifevision.add(actions.gotbalance_happened.p)
+  end
 end
 
 if svo.haveskillset('anathema') then
-function svo.valid.usedanathemabalance()
-  svo.checkaction(svo.dict.stolebalance.happened, true)
-  lifevision.add(actions.stolebalance_happened.p, nil, 'anathema')
+  function svo.valid.usedanathemabalance()
```
TheLastDarkthorne, line 4681: Another comment so I know where to look when I get back at this, is all of this space formatting ok or something messed up? It looks fine and I'm not receiving mudlet errors so I guess nothing iffy happening here

svo (setup, misc, empty, funnies, dor).xml
```diff
@@ -22,7 +22,7 @@ svo.modules_version = svo.modules_version or {}
 svo.modules_version["svo (setup, misc, empty, funnies, dor)"] = 1
 svo.loader.setup = function()
 
-svo.version = "54"
+svo.version = "55"
```
TheLastDarkthorne, line 25: Another one for me. This should be 54 still since latest version is 53 and 54 was reverted

svo (install me in module manager).xml
keneanung: Do you mean this is something I break during reverting the reverts or something you did during your changes?

svo (trigger functions).xml
keneanung: It's fine and actually better than before as this function declaration is only done if the anathema skillset is available.

svo (setup, misc, empty, funnies, dor).xml
keneanung: For transparency sake (and if people already installed 54) you should use 55 here.

svo (install me in module manager).xml
TheLastDarkthorne: Oh it is from the original author from the original PRs I reverted

keneanung:
```suggestion
  local newclass = (override ~= "gmcp.Char.Status") and override or gmcp.Char.Status.class
```

TheLastDarkthorne:
Seems good, I tested a few and no errors so far from the changes and nothing odd happening that I could see. Gonna merge it now

PR #808, opened 2023-03-07
Fix missing paladin debug message and ndb checking for enemies

### LINKED ISSUE #807: Capturing city enemies needs updating
I was helping someone fix their triggers for city enemy capture since conquest has changed the actual names of the cities to no longer necessarily have 'City' in the name. I will fiddle with it when I learn what the other ones are and make sure I can actually get it working. 
For the record, Hashan's is no longer the City of Hashan, it is                      Enemies of the Starforged Nation of Hashan:

Eleusis' hasn't changed yet, and I can only guess at the others or actually run around and look at each of them. Either way I know the script will need to be re-worked slightly since it's assuming everything not city or order is a house.

TheLastDarkthorne:
Missed a debug message in the paladin balance stuff. Also fixes #807

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -967,18 +967,6 @@ disableTrigger("NameDB qw")</script>
 				<Trigger isActive="no" isFolder="no" isTempTrigger="no" isMultiline="yes" isPerlSlashGOption="no" isColorizerTrigger="no" isFilterTrigger="no" isSoundTrigger="no" isColorTrigger="yes" isColorTriggerFg="no" isColorTriggerBg="no">
 					<name>City/House/Order enemies</name>
 					<script>local field
---[[
-if multimatches[2][1]:find("City") then
-  ndb.fixed_set(ndb.db.people.cityenemy, 0)
-  field = "city"
-elseif multimatches[2][1]:find("Order") then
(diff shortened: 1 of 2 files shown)


svo (namedb).xml
```diff
@@ -1892,7 +1898,6 @@ temp_name_list[1].mark = "Ivory"</script>
 						<regexCodePropertyList>
 							<integer>1</integer>
 							<integer>1</integer>
-							<integer>-1509852592</integer>
```
TheLastDarkthorne, line 1895: @vadi2 or @keneanung can you explain those integers? I have no idea what they are and why they are getting modified here, I only literally changed code lines and not messed with any triggers settings...

vadi2: It looks like an error to me. Is it always changing when you edit it?

TheLastDarkthorne: Looks like it... every time I rollback to the previous version and then edit and save the trigger it is doing that. Maybe it was some leftover from older mudlet versions?

vadi2: this is on 4.16?

could you show a screenshot of what pattern type are you setting that to?

TheLastDarkthorne: image
This is how it looks like (I didn't change any of them, only the code... well except for the first one that I adjusted just now)

PR #810, opened 2023-03-23
Possible fix for prios resetting and random moveWindow errors

TheLastDarkthorne:
Possible fix for prios resetting on class change or randomly through loading. It has been a bug for a long time and after some brainstorming and huge help from Irimon we think that this function was attempting to export at the same time svo was resetting prios to default as it loaded again which could be the reason why it randomly reset to default cause it was saving/loading the default configs again and again. Adding it to pre emit will ensure the export will always run first although it requires more testing to be sure it works.

Also includes a fix for the random moveWindow errors causing by createLabel function svo was attempting to use internally.

### DIFF
--- a/svo (install, config, pipes, rift, parry, prios).xml
+++ b/svo (install, config, pipes, rift, parry, prios).xml
@@ -3963,11 +3963,13 @@ end
 
 do
   local oldCL = createLabel
+  --[[
   function createLabel(name, posX, posY, width, height, fillBackground)
     oldCL(name, 0, 0, 0, 0, fillBackground)
     moveWindow(name, posX, posY)
     resizeWindow(name, width, height)
   end
+  ]]
 end
 
 function svo.toggle_riftlabel(toggle)
@@ -4958,7 +4960,13 @@ function prio.import(name, echoback, report_errors, use_default)
   return true
 end
 
+--[[
 signals.saveconfig:connect(function ()
+  prio.export('current')
+end, 'save prios on quit')
+]]
+
+signals.saveconfig:add_pre_emit(function ()
   prio.export('current')
 end, 'save prios on quit')
 
--- a/svo (install, config, pipes, rift, parry, prios).xml
+++ b/svo (install, config, pipes, rift, parry, prios).xml
@@ -3963,13 +3963,6 @@ end
 
 do
   local oldCL = createLabel
-  --[[
-  function createLabel(name, posX, posY, width, height, fillBackground)
-    oldCL(name, 0, 0, 0, 0, fillBackground)
-    moveWindow(name, posX, posY)
-    resizeWindow(name, width, height)
-  end
-  ]]
 end
 
 function svo.toggle_riftlabel(toggle)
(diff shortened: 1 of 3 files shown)


keneanung (approved):

vadi2:
Sorry, I don't follow the prio explanation - mind walking me through it with examples in the code?

svo (install, config, pipes, rift, parry, prios).xml
```diff
@@ -3963,11 +3963,6 @@ end
 
 do
   local oldCL = createLabel
```
DazedxJoker, line 3965: Do we still need this line if we aren't overwriting the function?

TheLastDarkthorne:
> Sorry, I don't follow the prio explanation - mind walking me through it with examples in the code?

In the classchange function there is:
```lua
    svo.signals.saveconfig:emit()
        
        -- reset all serverside prios so serverside doesn't spam class-specific defences
        -- it's a bug in the game that serverside still tries to put up defences you can't
    svo.sk.resetserversideprios()
        
    svo.systemloaded = false; svo_init_system()
```
Apparently, while that signal is being emitted and the connected functions executed, the system is reloading right at the same time too. The problem appears to be that the prio.export() function is being executed right at the same time as all this happening, which is apparently making it export and saving the prios while svo is either resetting the prios on reloading OR rebuilding the dict, causing it to wrongly save a jumbled up list of prios instead of the proper one.
The resulting list of prios when this bug happens is actually **different** than the default prios really, which makes me believe even more about the 'trying to save prios while rebuilding dict' theory, but... I unfortunately lack knowledge to properly pinpoint where on the code this is happening, but after talking with Irimon for a while we came up with this idea to make the prio.export() function always run first before all the others connected to the signal and it seems to be working so far...

But since this bug happens randomly I will keep it awhile before merging and ask some people to test too.

vadi2:
> Apparently, while that signal is being emitted and the connected functions executed, the system is reloading right at the same time too. 

That's not possible though, Lua VM is single threaded and everything you see here happens one line after another. Two things cannot be running at the same time.

So I'm not certain if this particular theory. Verify it by printing echoes right at the start and before the end of the relevant functions to get more insight here. Let's also do a lot of testing it see if the issue is really fixed, because the hypothesis we have right now doesn't really have a leg to stand on?

TheLastDarkthorne:
> > Apparently, while that signal is being emitted and the connected functions executed, the system is reloading right at the same time too.
> 
> That's not possible though, Lua VM is single threaded and everything you see here happens one line after another. Two things cannot be running at the same time.
> 
> So I'm not certain if this particular theory. Verify it by printing echoes right at the start and before the end of the relevant functions to get more insight here. Let's also do a lot of testing it see if the issue is really fixed, because the hypothesis we have right now doesn't really have a leg to stand on?

Looks like the dict was also being rewritten randomly (specially if you are navigating the scripts in mudlet and then click the dictionary script, which was causing the prios to get messed up somehow). I will test later to see the order of the funcs connected to that signal to see how they are being called

svo (install, config, pipes, rift, parry, prios).xml
TheLastDarkthorne: Hmmm I don't know.... there is a 'do' there and I have no idea where the statement for it starts and I'm a bit afraid of removing it and breaking the whole structure '-'

vadi2:
If the signal fix has been working fine, we can merge it in too.

Issue #809, opened 2023-03-20
NDB Undead faction + proposed fix


SlyViolin:
Morning. This is my directions as a proposed fix to the undead faction. I am not confident as tweaking the main body of the svof code so I leave this to my capable hands
------------------------------------------------------------------------------
TRIGGER NAME : NameDB qw
Find and replace

local city_colours = {
  ["0,255,0"] = "Eleusis",
  ["0,128,128"] = "Cyrene",
  ["255,0,0"] = "Mhaldor",
  ["128,128,0"] = "Hashan",
  ["128,0,128"] = "Ashtan",
  ["255,255,255"] = "Targossas",
  ["128,0,0"] = "Undead"
}
-----------------------------------------------------------------------------------
SCRIPT NAME : ndb.valid
add "Undead" to ndb.valid.cities

ndb.valid.cities = {
  "Ashtan", "Cyrene", "Hashan", "Eleusis", "Mhaldor", "Targossas", "Undead"
}
-------------------------------------------------------------------------------------
SCRIPT NAME : ndb main code
 add this line to local singular_city, local plural_city

Undead = 'Undead',

PR #812, opened 2023-03-29
adding undead to to qwc trigger and ndb

tsartin:
adding undead to to qwc trigger and ndb. Fixing minor capitalization error. proposed in #809

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -764,7 +764,7 @@ for _, name in ipairs(data) do
     if city_colours[rgb] then
       temp_name_list[#temp_name_list].city = city_colours[rgb]
     else
-      temp_name_list[#temp_name_list].city = "Rogue"
+      temp_name_list[#temp_name_list].city = "rogue"
     end
     setFgColor(192,192,192)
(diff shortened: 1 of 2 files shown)


svo (namedb).xml
```diff
@@ -763,7 +764,7 @@ for _, name in ipairs(data) do
     if city_colours[rgb] then
       temp_name_list[#temp_name_list].city = city_colours[rgb]
     else
-      temp_name_list[#temp_name_list].city = "rogue"
+      temp_name_list[#temp_name_list].city = "Rogue"
```
TheLastDarkthorne: Apparently this is by design as ndb has it lowercase as well. At least it does not seem to be breaking highlighting and recognizing of rogue people

tsartin: Think I should change it back?

TheLastDarkthorne: yeah should, unless you can confirm it is not breaking ndb on rogues

TheLastDarkthorne:
Just taking advantage to add a few things I noticed were missing!

PR #819, opened 2023-04-09
Fixes / QoL for Undead

MediaResAchaea:
NDB's manual honors does a request to the API for the information available there. The Undead's city is called 'underworld' there, so this gets persisted improperly after Honors, and until you correct it with a QWC, at which point the color registration fixes things. A potentially 'better' fix would be to just coalesce towards Underworld and update the QWC naming instead. 

I believe this discrepancy may also have been causing the issues with ndb.isenemy not working consistently on Undead.

Also adds an ndb.isundead function, identical to the existing ndb.is<city> functions.

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -3235,6 +3235,10 @@ function ndb.downloaded_file(_, filename)
     if city ~= "(hidden)" then
         temp_name_list[1].city = (city ~= "(none)" and string.title(city) or "rogue")
     end
+    -- API returns 'underworld' for the Undead.
+    if city == 'underworld' then
+        temp_name_list[1].city = 'Undead'
+    end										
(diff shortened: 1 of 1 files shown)


svo (namedb).xml
```diff
@@ -3235,6 +3235,10 @@ function ndb.downloaded_file(_, filename)
     if city ~= "(hidden)" then
         temp_name_list[1].city = (city ~= "(none)" and string.title(city) or "rogue")
     end
+    -- API returns 'underworld' for the Undead.
+    if city == 'underworld' then
```
keneanung, line 3239: This will overwrite the city value twice (once for the if clause above, once here). It's fine in this case, I just wanted to point it out as a learning opportunity :wink:

keneanung (approved):

PR #821, opened 2023-04-15
Revert regex for issue #820

### LINKED ISSUE #820: Over-inclusive regex causes NDB mis-assignment
https://github.com/svof/svof/blob/683b187de27a6fcfebba59596ce1aa3b89660bd2/svo%20(namedb).xml#L1763

This regex change introduced a minor issue as it no longer only catches only on city, and instead captures 'the army of X' where X can be any city. 

Given we don't actually need to rely on this to identify undead, I'd suggest just reverting the regex?

MediaResAchaea:
This regex change was overly greedy, and causes us to assign people to the city of "the army of X" until qwc fixes it.

We could also just make the regex better, but I don't think capturing whether or not an undead is in the undead army is worthwhile?

Fixes #820

### DIFF
--- a/svo (namedb).xml
+++ b/svo (namedb).xml
@@ -1760,7 +1760,7 @@ temp_name_list[1].city = matches[3]</script>
 						<colorTriggerFgColor>#000000</colorTriggerFgColor>
 						<colorTriggerBgColor>#000000</colorTriggerBgColor>
 						<regexCodeList>
-							<string>^\w+ is an? (.+) in (.+)\.$</string>
+							<string>^\w+ is an? (\w+) in (\w+)\.$</string>
 						</regexCodeList>
 						<regexCodePropertyList>
 							<integer>1</integer>


tsartin:
Would it be beneficial to know if they are in the army during raids? 

I can't think of any reason if not this.

TheLastDarkthorne:
I just wanted to capture it so people can know which rank they are. But I suppose their ranks are also tied to their titles as well. I should have thought of a better regex lol
But it is fine... It was just for fancy, I don't think we really need to know their army rank specially since it seems to be also tied to their title and progress in undeath skill.

MediaResAchaea:
Undead rank in army doesn't actually change with the elevated level in their title, they're always just:

"a Profane Warrior in The underworld."

And all undead are soldiers, so there's no actual information provided from this line.

MediaResAchaea:
I get the intent though, and less greedy (or more scoped) regex would be a "better" fix here for the future!

TheLastDarkthorne:
I had no idea lol I thought it changed
Thank you really, I will push this soon

Issue #813, opened 2023-03-30
Update Lock Data


tsartin:
Add sandfever to svo.cnrl.lockdata anywhere where impatience is mentioned. 

With the addition of new afflictions there might be other afflictions that need to be added here.

tsartin:
Might consider adding affs that knock off focus balance too. It can be used as a semi-lock

TheLastDarkthorne:
> Add sandfever to svo.cnrl.lockdata anywhere where impatience is mentioned.
> 
> With the addition of new afflictions there might be other afflictions that need to be added here.

Why sandfever? And yeah I got some advice in the achaea discord channel so I will add those but need more explanation on sandfever. Does it causes a lock or it just have an interaction with impatience?

> Might consider adding affs that knock off focus balance too. It can be used as a semi-lock

I have no idea, I don't fight in Achaea anymore for a long time, I don't even know what the new affs do lol But I suppose this function is to warn when they are effectively lock, but could add a warning for those knock off focus affs too maybe

MediaResAchaea:
@TheLastDarkthorne Sandfever blocks Focus, the same as impatience, so for locking purposes they're interchangeable.

PR #823, opened 2023-04-21
Focus, Locks, and Stain updates

MediaResAchaea:
Well, this started as a five second attempt to update the lock data for #813 and then spiraled a bit further. From the discussion that @TheLastDarkthorne kicked off in the Achaea discord (https://discord.com/channels/872599744793419776/874411546141417492/1098763812791914518) it was pointed out that the entire concept of a Stain lock doesn't exist anymore. 

In fact, Stain hasn't blocked any cures for 7 years, and the logic within Svo was never updated to reflect this.

This PR does 4 things:

1) Updates svo.cnrl.lockdata to remove Stain locks, and updates every warning to consider Sandfever equivalent to impatience for locking purposes, as Sandfever fully blocks focus.

2) Updates the Focus illusion check and check_focus to treat sandfever as equivalent to impatience

3) Removes all traces of Stain's impact on curing Slickness, resolving 7 years of bad logic that made it so Svo wouldn't try to cure Slickness if stain was active.

4) Updates the documentation to reflect the existence of the Dragon lock warning, and make the descriptions consistent. It's not actually clear to me if this page is actually served anywhere from the in-client branch, but no harm either way I guess.

Warning: This is my first time ever touching this stuff, please validate I didn't do anything silly! They're all pretty straightforward changes, though, so hopefully it's fine!


svo (setup, misc, empty, funnies, dor).xml
MediaResAchaea, line 1: Oh, this actually isn't quite right.

MediaResAchaea: Should be right now, but would appreciate vetting!

svo (setup, misc, empty, funnies, dor).xml
```diff
@@ -2840,9 +2840,7 @@ end
 
 empty.eat_bloodroot = function()
   svo.lostbal_herb()
-  svo.rmaff('paralysis')
-
-  if not affs.stain then svo.rmaff('slickness') end
+  svo.rmaff('paralysis', 'slickness')
```
TheLastDarkthorne: rmaff function only accepts one arg
just call it twice once for para and once for slick

TheLastDarkthorne:
Sounds all good, just fix that one little thing and make sure it is all working (like focusing properly removing the affs if you got them and the lock warnings working with the sandfever changes, you can use `svo.addaff` to make svo think you have an aff)

svo (setup, misc, empty, funnies, dor).xml
Sebguer: oh, rip, i thought i copied this from somewhere but i guess it was a different function 🤦‍♂️

MediaResAchaea: I updated this to an array of the affs, which I think is right based on existing (which is what I was trying to copy but missed that it should be an array):

https://github.com/MediaResAchaea/svof/blob/ae640b2d4ea8214b84dc26d44558c2023486fe9d/svo%20(setup%2C%20misc%2C%20empty%2C%20funnies%2C%20dor).xml#L2921

MediaResAchaea:
Testing log: https://ada-young.com/pastebin/HLtiJKoC

Everything looks to work as expected.

(Note: It focuses through Sandfever in the opening bit because I had forgotten to pull over the focus change when I was testing the lock warnings 😅)

MediaResAchaea:
Not sure how I missed these, but there were also default scenarios for when to use shrugging, restore, and dragonheal to escape a stainlock.

TheLastDarkthorne:
Really good catches here, maybe you will understand the system better than I do lol I totally overlooked that you can pass a table to the affliction functions and didn't see the scenarios thing either.

Tbh I would never thought of the stain stuff until you pointed it out lol

PR #734, opened 2021-08-04
Gmcp tracking

awitteio:
Allows proper tracking of class resources from gmcp to svof.

- Created a new location to store class related info
- Altered back-end stuff to relate to the new location of class info
- Added in each classes proper tracking 

1. Alchemist (Wellspring)
2. Apostate (Essence)
3. Bard (Voice Balance, Stance)
4. Blademaster (Shin, Stance)
5. Depthswalker (Word Balance, Age)
6. Druid (Sunlight, Morph)
7. Infernal (Essence, Spec, Ferocity, Momentum)
8. Magi (Elementalism buffs)
9. Monk (Path, Stance, Kata, Kai)
10. Occultist (Karma, Entity)
11. Paladin (Devotion, Spec, Ferocity, Momentum)
12. Pariah (Epitaph Length)
13. Priest (Devotion, Conviction, Prayer Length, Prayer Balance)
14. Runewarden (Spec, Ferocity, Momentum)
15. Sentinel (Morph)
16. Serpent (Secreted Venom)
17. Shaman (Spirits)
18. Sylvan (Sunlight, Weatherweaving Buffs)
19. EartLord (Shaping Count)
20. Fire (Spark Count)

PR #825, opened 2023-04-26
GMCP resource tracking 2.0

TheLastDarkthorne:
Based on Kriex's PR #734 with updated gmcp tracking for all classes, but retaining original variable locations due to issues with that approach. Since he's been MIA I decided to create another PR based on his awesome additions!

Work in progress so this is not ready yet, but creating the PR so I don't lose my changes. There are things that need to be done like:

- [ ] Delete triggers that are being used to populate the resources
- [ ] Add prompt tags to all other things (or consolidate them all into one tag e.g. `@class` or whatever)
- [ ] Delete anything that is capturing the values through prompt flags if any
- [ ] Test of course!

This is a fairly large change so please if someone can look at it and see if it is all okay and/or is able to test through multiple classes I will be thankful.

PS: Apparently, `stats` or `svo.stats` is used for quantified stats (numbers, like sunlight, essence, devotion, etc) and `me` or `svo.me` is used for the rest of things so I tried to retain that convention. There should be probably a comment explaining this but I can place one later too.


MediaResAchaea:
I use svo.stats and svo.me in a bunch of places, so let me know whenever this is ready for testing and happy to make sure everything is working!

PR #830, opened 2024-05-23
fix: SVOF should now correctly have two events to set server side `mosshealth` and `mossmana`

veles-achaea:


vadi2 (approved):

DazedxJoker (approved):
Looks good!

PR #831, opened 2024-05-23
fix: fixing assess to correctly display for fae

veles-achaea:
Quick patch for assess to correctly cater for faes health.


vadi2 (approved):

DazedxJoker (approved):
Looks good to me

PR #832, opened 2024-06-04
feat: supporting mounting of urns

veles-achaea:
Supporting new game syntax such as `vault urn/veles` for `mount`ing and `vault`ing, for the Shop of Wonders `A petite ceramic urn`.


vadi2 (approved):

ndsamuelson (approved):
can confirm it does work. and you learn something new every day. hah.

PR #833, opened 2024-06-04
Pyre affliction from Excision (Paladin)

veles-achaea:
Putting up an initial pull request to get the conversation going, initial testing seems to show that this is working fine, but I know there are a few kinks that I couldn't work out, and figured it would be best to put it out there.

The main things that are on my mind/that I noticed:
- [ ] Aff doesn't appear to be serverside enabled by default, even though serverside is enabled - is there a way to jerry rig this for those upgrading?
- [ ] I'm not sure if I got the changes right in the `default_prios` file as it failed to import and required another patch (which is not included here - `name` being `nil` which could also be a Mac thing)


ndsamuelson:
Beat me to it. Good man. Testing but it looks the same as what I did to solve it so it's Colgy supported.

ndsamuelson (approved):
I'm not getting the previous errors with the changes in place. You managed to fix it where I managed to completely break mine when I was working on it. Cheers.

svo (actions dictionary).xml
```diff
@@ -1772,6 +1772,32 @@ if not next(svo.dict) then
         oncompleted = function() svo.rmaff('justice') end,
       }
     },
+    pyre = {
+      herb = {
+        isadvisable = function()
+          return (affs.pyre and (not (svo.haveskillset('discipline') or svo.haveskillset('malignity') or svo.haveskillset('valour') or svo.haveskillset('domi
```
vadi2: `'domi` is split across two lines by accident, if you bring it back on one line, the syntax error will go away.

> Error in Script 'Dictionary of actions (affs/defs/misc)': [string "-- Svof (c) 2011-2018 by Vadim Peretokin..."]:1767: unfinished string near ''domi'

ndsamuelson: I was just looking at that and wondering why. You're the man.

veles-achaea: Good pickup, thought I was losing my mind. Thank you!

veles-achaea:
Fixed the split string issue!

vadi2:
:+1: I am not familiar with how the mechanic is supposed to work - could you add a description https://github.com/svof/svof/pull/833#issue-2332821009?

ndsamuelson (approved):
can confirm the changes solved the syntax check issue. 👍

veles-achaea:
I don't have Paladin myself, and this should probably be tested in more depth, but from the Wiki:

```
Pyre | Consign them to the flame.
-- | --
Syntax: | PERFORM PYRE <target>
Required: | 5% Holy Wrath
Cooldown: | 2.5 seconds of wrath balance
Details: | Begin consigning a heretic to the flame, purging Creation of their blight

(*).Gives the pyre affliction. This makes it impossible to cure ablaze below the level of the pyre (so
if you have level 3 pyre, you cannot cure ablaze lower than level 3).
```

From what I saw, it definitely also sacks beyond 1.

veles-achaea:
Is there a way to get it to block curing blazing on client side? I actually only use server side. I'm also unsure about the rage and other ability checks, if you could add any guidance, that would be great!

svo (actions dictionary).xml
```diff
@@ -1772,6 +1772,31 @@ if not next(svo.dict) then
         oncompleted = function() svo.rmaff('justice') end,
       }
     },
+    pyre = {
+      herb = {
+        isadvisable = function()
+          return (affs.pyre and (not (svo.haveskillset('discipline') or svo.haveskillset('malignity') or svo.haveskillset('valour') or svo.haveskillset('dominion')) or not svo.dict.rage.misc.isadvisable())) or false
```
vadi2, line 1778: Not 100% sure here but will it work for classes that don't have rage?

vadi2:
> Is there a way to get it to block curing blazing on client side?

Yeah, adjust the isadvisable for blazing affs to check the level of pyre

svo (actions dictionary).xml
TheLastDarkthorne: This is strange, I don't understand those isadvisable checks. Why is it checking if the user has either of those knight skills? And uh.. Rage? I don't think the affliction only affects knights or interferes with rage? Either this one or the justice above? My guess is that whoever wrote this before mistook if for a skill maybe?

Im probably being stupid lol

Edit: nvm it is checking for knight rage. I forgot what this ability does. Heal affs?

veles-achaea: Aye, it's just my bad copy pasta! Was chatting to Vadi and it's there to see if rage will heal it.

I've asked on the main disc to see if this is healable by rage, and will also need to add in the checks for ablaze to ensure that it can't be cured <= 3 while pyre is on the target.

vadi2: @TheLastDarkthorne I didn't remember what it did at first too, but in justice and other rage-healable affs it's just a safety check to ensure that both herb and rage cures don't go off at the same time. Like @veles-achaea mentions we need to verify if rage can cure this aff too

veles-achaea: Okay, @ndsamuelson confirmed with Puxi that rage does in fact cure it. So that's right to leave I believe, I still don't understand the full checks as some are checking for skills, while the other is checking the rage table.

So I need to add it to the rage table as well to close this out, and restrict curing ablaze, and I think we should be good to close this one out?

veles-achaea: Any idea on a way to get the prios in for people? Or will they need to manually fidget with this, and also do something like `tn affs` to also disable client side handling of it?

veles-achaea:
Note to self, needs diag trigger as well, and tree cured lines, and general cures? Search for `ablaze`

svo (actions dictionary).xml
vadi2: There's a few things that could be done here, but what are you looking to do exactly? Remind people to reset their prios to default for the new thing?

veles-achaea:
Does anyone know how to handle the various levels on this one? It looks like ablaze can stack, as well as pyre, and ablaze must be <= pyre from what I can gather. Also, there are a lot of triggers to pick up the affliction for ablaze, but is the same thing required for pyre? I can't seem to find any lines for justice, unsure if at some point SVOF was flipped to pick up GMCP affs instead of lines?

Keen to close this one out, just not sure how to best do so.

vadi2:
GMCP tracking is in: https://github.com/svof/svof/pull/209

For ablaze, check in pyre isadisable that ablaze level is lower than pyre, that should do it.

vadi2:
That was an accident

veles-achaea:
How do you check the levels for pyre and ablaze though? And is something needed to configure that they are afflictions that stack?

vadi2:
Ablaze was done as separate afflictions (I think because they were called so in the game), so you'd have to check the individual pyre level against the individual ablaze aff. Pyre itself could be implemented as a stackable affliction though, have a look at `skullfractures` for inspiration!

TheLastDarkthorne:
> Ablaze was done as separate afflictions (I think because they were called so in the game), so you'd have to check the individual pyre level against the individual ablaze aff. Pyre itself could be implemented as a stackable affliction though, have a look at `skullfractures` for inspiration!

I think all stackable affs now has a count property in gmcp (not entirely sure about this as I haven't looked in a long time, but some quick tests could tell maybe). Might be a good idea to clean all these burning shenanigans and just use that instead in the future as it is not really needed anymore, like how unweavings and other stackable stuff are tracked through gmcp as well.

vadi2:
That's a good idea but it might break scripts and any prios that gets shared around. Maybe utility functions to help you deal with stacking will go further without breaking things 🫣

@veles-achaea have you got an idea of what to do or would an example here help?

veles-achaea:
Hi team, apologies, been back at work and struggling to find the time to close the stacking and client side out. Happy for someone else to pick up if possible, I know there are going to be plenty of upcoming Bard changes with their rework too, would be good to get this in before that to not have to come back to it.

vadi2:
Thanks for your help with this!

Is anyone else able to pick this up? Otherwise we should merge as-is since it is a good improvement already.

TheLastDarkthorne:
> Thanks for your help with this!
> 
> Is anyone else able to pick this up? Otherwise we should merge as-is since it is a good improvement already.

I can try to do this. I suppose it is just inserting a check for pyre level at each of the burning affs `isadvisable` property?

vadi2:
Yep!

veles-achaea:
Aye, that and if anything needs to be done to make it a stackable aff which it is. Thank you so much for assisting!

TheLastDarkthorne:
> I don't have Paladin myself, and this should probably be tested in more depth, but from the Wiki:
> 
> ```
> Pyre | Consign them to the flame.
> -- | --
> Syntax: | PERFORM PYRE <target>
> Required: | 5% Holy Wrath
> Cooldown: | 2.5 seconds of wrath balance
> Details: | Begin consigning a heretic to the flame, purging Creation of their blight
> 
> (*).Gives the pyre affliction. This makes it impossible to cure ablaze below the level of the pyre (so
> if you have level 3 pyre, you cannot cure ablaze lower than level 3).
> ```
> 
> From what I saw, it definitely also sacks beyond 1.

I have converted the aff into stackable aff, but I need to do the logic on the burning affs still
The skill says that it makes impossible to cure ablaze below the level of the pyre, but what about the initial pyre affliction? The first stack I mean, does it make impossible to cure ablaze until it is cured as well?

vadi2 (approved):
Merging now so we can keep working on it

PR #834, opened 2024-06-25
Fleshed out pyre conditions and entry

TheLastDarkthorne:
Continuing the work from #833. Changed pyre to be a stackable aff and changed the conditions in the burning affs so that it doesn't cure when pyre affs are above their threshold level. I also removed some mentions to old cures that are not being used anymore, as well as a fix in the horror aff entry I caught.

This should work but would be good to have some testing. I also don't know about the `burning` entry though, maybe a remnant of when it was trigger based to recognize catching ablaze from the environment/room?

### DIFF
--- a/svo (actions dictionary).xml
+++ b/svo (actions dictionary).xml
@@ -1089,7 +1089,6 @@ if not next(svo.dict) then
         end,
   
         noeffect = function()
-  			  --svo.rmaff('horror')
   			  svo.lostbal_herb()
   		  end,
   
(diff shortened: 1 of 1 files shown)


svo (actions dictionary).xml
```diff
@@ -1089,7 +1089,7 @@ if not next(svo.dict) then
         end,
   
         noeffect = function()
-  			  svo.rmaff('horror')
+  			  --svo.rmaff('horror')
```
vadi2: better to delete it or add a comment, otherwise when someone else comes across this in the future they will not know why is it here commented out

vadi2 (commented):
Looks awesome! Lets get it tested out.

vadi2 (approved):

TheLastDarkthorne:
Ok gonna merge this! Sorry for the delay!

PR #836, opened 2024-09-30
Hopefully fixed conditions for pyre

TheLastDarkthorne:
Conditions for pyre check was apparently wrong and erroring when people got ablaze and didn't have pyre. Tested this with a few people and on my own system and it seems to work now.

Still couldn't test it on battle against a Paladin though... but hopefully it will work.


vadi2 (approved):

TheLastDarkthorne:
Ok I will merge this cause it seems to be working fine, at least with my tests anyway. I have asked other people to test but didn't get a response

PR #837, opened 2025-02-27
Namedb and curing priority queue improvements

TheLastDarkthorne:
Added non-binary capture support for ndb and an improvement to help mitigate problems with curing priority limits

### DIFF
--- a/svo (curing skeleton, controllers, action system).xml
+++ b/svo (curing skeleton, controllers, action system).xml
@@ -2272,7 +2272,7 @@ function sk.dosendprios()
     if #sk.affqueue &gt; 0 then
         local what = "curing priority"..table.concat(sk.affqueue)
         if sk.queuecount &gt; 4 then
-          tempTimer(1,function()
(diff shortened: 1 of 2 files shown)


svo (curing skeleton, controllers, action system).xml
```diff
@@ -2272,7 +2272,7 @@ function sk.dosendprios()
     if #sk.affqueue &gt; 0 then
         local what = "curing priority"..table.concat(sk.affqueue)
         if sk.queuecount &gt; 4 then
-          tempTimer(1,function()
+          tempTimer(1+getNetworkLatency,function()
```
vadi2: I think you need a () here and in others

vadi2 (approved):
Nice job 🙂

svo (curing skeleton, controllers, action system).xml
TheLastDarkthorne: oops my bad o.o gonna fix it

PR #839, opened 2025-03-27
Add new affs, frost fix

TheLastDarkthorne:
Adding new affs that haven't been added yet as well as fixing something weird with frost being marked as undeffable when it actually can be handled by serverside


vadi2:
Trying something new...

vadi2 (approved):
Lgtm

TheLastDarkthorne:
Found out that the oniar arena wasn't added in the valid arena entries so I added it, was causing issues with the cure balance tracking inside the arena

vadi2:
Good catch!

Feel free to merge whenever ready.

PR #842, opened 2025-11-06
Add fixes for pyre and additions to gmcp tracking and some other stuff

TheLastDarkthorne:
Work in progress so there's some debug stuff in the code I will remove later! Have to test some stuff still. Paladin pyre was completely breaking the system, so a fix was long overdue. For some reason affs.pyre.count gets nil or don't get populated at the same time as we get affliction so I had to come with an workaround, thanks to Ievandur for the idea!
Also added a way for the system to update affliction stack count upon receiving info from gmcp affliction list. I don't know how well it is working but so far I have not encountered issues.

### DIFF
--- a/svo (actions dictionary).xml
+++ b/svo (actions dictionary).xml
@@ -1771,6 +1771,7 @@ if not next(svo.dict) then
         oncompleted = function() svo.rmaff('justice') end,
       }
     },
+    
     pyre = {
       count = 0,
       herb = {
@@ -1825,6 +1826,7 @@ if not next(svo.dict) then
         end,
       }
     },
+    
     generosity = {
       herb = {
         isadvisable = function()
@@ -4514,7 +4516,7 @@ if not next(svo.dict) then
         noeffect = function() svo.rmaff('mildconcussion') end
       }
     },
-  
+
   
   -- salve cures
     anorexia = {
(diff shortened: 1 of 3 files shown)


TheLastDarkthorne:
Closing this. The burn-entry half is superseded by #853, and the GMCP half is superseded by #846/#847 - but two of the trigger changes here found real bugs that are still live, noted below so they are not lost with the PR.

### GMCP List handler - superseded, and the implementation here does not work

All three ideas are already on `in-client-svof`, done differently:

- `tonumber(string.match(name, "%d"))` takes the first digit anywhere in the string. That is the exact misparse #846 replaced with `parseaffname`, matching `^(.-) %((%d+)%)$` - the old form breaks at level 10 and on any affliction name that merely contains a digit.
- `if thisaff:sub(-4) == " (%d)"` cannot ever be true. `sub` returns a literal substring, `" (%d)"` is a literal five-character string, and `sub(-4)` yields four characters. So the suffix is never stripped, `sstosvoa[thisaff]` stops resolving, and this breaks even the level-1 case that worked before.
- Writing the level into `svo.dict[x].count` is the right idea and landed in #847, with `svo.updateaffcount()` alongside it so `svo.affl` and the `svo updated aff` event follow. Setting the dict alone leaves everything downstream reading the old value.

It also leaves the `ipairs(svo.affl)` bug in place - `svo.affl` is string-keyed, so `ipairs` iterates nothing, `preaffl` stays empty and the removal loop below is dead. #847 fixed that with `pairs`. And there are two `display(thisaff)` calls.

### Burn entries - superseded by #853

The mechanics quoted in the comment here are right and are what #853 is built on, credited there. The code needed three corrections: `affs.pyre.count` does not exist (the level is on `.p`; `(affs.pyre and affs.pyre.count) or 0` silences the raise but pins the level at 0, so the gate never engages), the thresholds were one rung too permissive at every level, and presence has to floor the level at 1 for anyone without GMCP affliction messages. #853 also adds a logic test for the ladder.

### Two things here that are still worth doing

Both verified against the current tree, neither carried into #853:

1. **The `blistered` diag trigger ships an empty script.** `svo.valid.diag_blistered` is defined - `blistered` is in the list in `Diag_trigger_functions.lua` - but it is called from nowhere, so DIAGNOSE never records blistered. Pre-existing; the 2018 xml has it empty too. The fix here is correct.
2. **There is no pyre diag trigger at all.** No `diag_pyre` anywhere in the tree, and pyre is not among the diag folder's children, so DIAGNOSE never records pyre either. Adding one is worthwhile - it pairs with #853, where a diag-detected pyre immediately holds ablaze correctly via the presence floor. The pattern proposed here wants confirming against the real diagnose line first.

### Not to carry over

- `Finished` losing `svo.valid.diagnose_end()` would break the end of every diagnose. It reads like an accidental swap with the blistered edit in the same hunk.
- `waitingforsileris.customwait` 10 -> 8 may well be right, but it is a timing claim that needs measuring in game rather than taking on faith.
- The trailing space on the `svo noeffect moss ` trigger name is cosmetic - nothing references that item by name, so it is not a silent no-op - and renaming an item is structural to the verification gate, so it is not worth doing on its own.

TheLastDarkthorne:
Also the gmcp fixes from the stack we pushed before should be better than what I was cooking here too anyway

